PR #58, opened 2026-06-16
feat(fleet-console): inject /rename into session PTY on Operation rename

sbluemin:
## Summary
When an Operation (terminal session) is renamed in the Fleet Console, the running Agent CLI is now kept in sync by injecting a `/rename <name>` slash command into that session's PTY.

This reuses the existing, validated **carrier-result reminder injection path** (`sanitizeCarrierResultReminder` + `formatCarrierResultReminderMessage` from `@dotobokuri/fleet-admiral`, written via `TerminalSessionManager.writeToSession`), so per-CLI `lineTerminator` / bracketed-paste handling stays consistent with carrier reminders. No new injection mechanism is introduced.

## Changes
- `runtime/fleet-console/src/server.ts` — new `injectRenameCommand` helper wired into the PATCH `/terminal/sessions/:id` rename handler (runs after the label update).
- `runtime/fleet-console/tests/server.test.ts` — regression test covering normal, malicious-control-char, empty, and control-char-only labels.
- `CHANGELOG.md` / `runtime/fleet-console/AGENTS.md` — documented the new behavior.

## Security (command injection)
- The label is single-lined (`[\r\n\t]` → space) and sanitized (ESC / BEL / C1 / bracketed-paste markers stripped) **before** the `/rename ` prefix is assembled, so a label cannot split the line into an additional command.
- Control-character-only labels (which sanitize to empty) are skipped — no bare `/rename` leaks.
- Independently audited by Sentinel: **PASS** (no Critical/High command-injection paths).

## QA
- `typecheck` · `test` (257 passed) · `build` all green, rebased on top of `canary`.

## Notes
- On profiles with `bracketedPaste: true` (Codex/Claude), `/rename` is paste-wrapped exactly like carrier reminders. Whether a given Agent CLI executes `/rename` as a slash command is up to that CLI.


sbluemin:

## Addressed
- **P2 (gate rename injection by CLI support)** — `runtime/fleet-console/src/server.ts` + `packages/fleet-admiral/src/agent-cli/*` — `/rename` injection is now gated by a per-CLI `renameCommand` capability. The capability is declared as the single source of truth on `AgentCliDefinition` and surfaced via `AgentCliMetadata`; the console resolves the session CLI's rename command and **skips injection when the CLI (or an unknown `cliId`) declares no rename command**.

## Validation
- `pnpm --filter @dotobokuri/fleet-admiral typecheck && test && build` — pass (15 tests, DTS + boundary guard OK)
- `pnpm --filter @dotobokuri/fleet-console typecheck && test && build` — pass (258 tests, incl. a new "skips rename injection for sessions without a rename-capable CLI" case)

## Notes
- One factual correction, with respect: the P2 premise that Codex CLI lacks `/rename` does not hold — **Codex CLI does support `/rename`** (verified by the repo owner), so the previous unconditional injection was already correct for both Claude and Codex sessions. The change is therefore not a today-bug fix but a **structural gate**: rename capability is now declared per-CLI, so any future CLI lacking a rename slash command is automatically skipped instead of receiving an unsupported command. `claude`, `claude-kimi`, and `codex` all declare `"/rename"`.

sbluemin:
@codex The follow-up review feedback has been addressed. Please re-review.

## Addressed
- **P2 (gate rename injection on the actual launched CLI)** — `packages/fleet-admiral/src/agent-cli/*` + `runtime/fleet-console/src/terminal/*` + `src/server.ts` — rename injection is now driven by the **resolved launch profile**, not the originally requested `cliId`. `renameCommand` moved from `AgentCliDefinition` to `AgentCliProfile` (next to `messagePolicy`), flows through `TerminalLaunchSpec` into the session, and is read via `getSessionRenameCommand` — the same source as `messagePolicy`. A `FLEET_TERMINAL_CMD` override returns a launch spec with no `renameCommand`, so `/rename` is skipped (e.g. `bash -l` no longer receives it). Added a regression test for the override / unsupported-CLI case.

## Validation
- `pnpm --filter @dotobokuri/fleet-admiral typecheck && test && build` — pass (15 tests)
- `pnpm --filter @dotobokuri/fleet-console typecheck && test && build` — pass (258 tests)

## Notes
- Good catch. The prior approach gated on the user-selected `cliId`, which diverged from `messagePolicy` (resolved profile) precisely in the override path. The capability now lives on the same resolved-profile source, so arbitrary overrides and unsupported CLIs are handled consistently.

PR #87, opened 2026-06-19
feat(console): gate Operation launch by CLI install/sign-in state

sbluemin:
## Summary
- Operation 생성 런치 메뉴(사이드바·캔버스 우클릭)에서 미설치 또는 미로그인 상태의 Agent CLI 항목을 비활성화하고, 비활성 사유("Not installed" / "Sign-in required")를 라벨로 표시합니다.
- 비활성 사유 정책을 공용 helper(`launchDisabledReason`)로 분리해 모든 런치 UI가 동일 게이트를 공유합니다(SSoT).
- 서버 세션 생성 경로(`/observer/theaters/:id/sessions`, folder-grant 세션)도 미설치/미로그인 CLI 요청을 409로 거부해 API 경계에서 정책을 강제합니다.
- 설치(detect)·로그인(model-auth) 상태 결합을 순수 함수 `combineAgentCliLaunchMetadata`로 분리하고, CLI detector를 주입 가능하게 하여 서버 테스트를 환경 비의존으로 유지합니다. `available`/`signedIn`은 boolean만 노출합니다(Token Boundary).

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console test` (414 passed)

## Changelog
- [x] `.changelog.d/operation-launch-cli-gating-changed.md` 포함 (`section: Changed`, `[fleet-console]`)

### DIFF
--- a/.changelog.d/operation-launch-cli-gating-changed.md
+++ b/.changelog.d/operation-launch-cli-gating-changed.md
@@ -1,0 +1,4 @@
+---
+section: Changed
+---
+- [fleet-console] The Operation launch menus now disable an Agent CLI choice when its CLI is not installed, or when a sign-in-gated model is not signed in, labeling each disabled choice with the reason, and the server rejects session-creation requests for those CLIs.
--- a/runtime/fleet-console/client/src/api.ts
+++ b/runtime/fleet-console/client/src/api.ts
@@ -275,10 +275,16 @@ function assertTerminalFolderListEntry(value: unknown, status: number): Terminal
 
 function assertAgentCliMetadata(value: unknown, status: number): AgentCliMetadata {
   const payload = value as Partial<AgentCliMetadata>;
-  if (!payload || typeof payload.id !== "string" || typeof payload.label !== "string") {
+  if (
+    !payload
+    || typeof payload.id !== "string"
+    || typeof payload.label !== "string"
+    || typeof payload.available !== "boolean"
+    || typeof payload.signedIn !== "boolean"
+  ) {
     throw new ApiError(status, "Invalid Agent CLI metadata response");
   }
-  return { id: payload.id, label: payload.label };
+  return { id: payload.id, label: payload.label, available: payload.available, signedIn: payload.signedIn };
 }
 
 function assertObservedTenant(value: unknown, status: number): ObservedTenant {
--- a/runtime/fleet-console/client/src/canvas/canvas-context-menu.tsx
+++ b/runtime/fleet-console/client/src/canvas/canvas-context-menu.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useRef } from "react";
 
-import { agentCliIcon } from "../components/operation-launch-menu.js";
+import { agentCliIcon, launchDisabledReason } from "../components/operation-launch-menu.js";
 import type { AgentCliMetadata, ConsoleState } from "../types.js";
 
 interface CanvasContextMenuProps {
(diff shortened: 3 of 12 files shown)


sbluemin:

## Addressed
- P2 (operation-launch-menu.tsx:111) — model-auth-store.ts: signInModel/signOutModel now refetch the Theater bootstrap so the launch-menu signedIn gate updates immediately instead of staying stale until a full reload.
- P2 (server.ts:682) — buildAgentCliLaunchMetadata now calls migrateLegacyAuthStore() before reading auth, matching /model-auth/state, so users whose Kimi/GLM key lives only in the legacy store are no longer misjudged as signed-out and wrongly rejected.

## Validation
- pnpm --filter @dotobokuri/fleet-console typecheck — pass
- pnpm --filter @dotobokuri/fleet-console test — pass (414)
- pnpm --filter @dotobokuri/fleet-console build — pass

## Notes
- No items declined; both P2 suggestions were applied.

sbluemin:
@codex The latest feedback has been addressed. Please re-review.

## Addressed
- P2 (server.ts:682) — Removed the migrateLegacyAuthStore() call from buildAgentCliLaunchMetadata. Since the sign-out refresh now calls /observer/theaters, an unconditional migrate there would resurrect a just-removed legacy key and re-open the gate for a provider the user signed out of. The gate now reads the current auth store only.

## Validation
- pnpm --filter @dotobokuri/fleet-console typecheck — pass
- pnpm --filter @dotobokuri/fleet-console test — pass (414)
- pnpm --filter @dotobokuri/fleet-console build — pass

## Notes
- This reverses the earlier migrate addition (prior P2). The two suggestions are mutually exclusive given the known fleet-infra/auth SSoT limitation (migrate copies without deleting the legacy original, and deleteApiKey only clears the current store). We prioritized honoring the user's explicit sign-out over auto-exposing a legacy-only key, which is the safer balance. Legacy-only keys are still surfaced by /model-auth/state's migrate on Settings entry; fixing the underlying resurrection is tracked as a separate follow-up, consistent with the project decision recorded for the Model Sign-in work.

sbluemin:
@codex The latest feedback has been addressed. Please re-review.

## Addressed
- P2 (server.ts:775 / agent-cli-detect.ts) — The default availability detector now resolves each binary with the same precedence as launch (resolveBinary: CLAUDE_BIN/CODEX_BIN override first, then PATH). Users who point an override at an off-PATH absolute path are no longer reported as unavailable, so the gate no longer rejects valid override launches with 409.
- P2 (operation-launch-menu.tsx:118) — Arrow-key roving focus now filters to `[role^='menuitem']:not(:disabled)`, so keyboard users can navigate past disabled (not-installed / not-signed-in) items to reach enabled CLIs.

## Validation
- pnpm --filter @dotobokuri/fleet-console typecheck — pass
- pnpm --filter @dotobokuri/fleet-console test — pass (414)
- pnpm --filter @dotobokuri/fleet-console build — pass

## Notes
- No items declined; both P2 suggestions were applied.

sbluemin:
@codex The latest feedback has been addressed. Please re-review.

## Addressed
- P2 (server.ts:775) — When FLEET_TERMINAL_CMD is set, the launch resolver ignores cliId and runs the operator override, so the gate now bypasses the install/sign-in check in that case instead of returning 409.

## Declined (with rationale)
- P2 (server.ts:686, legacy-only auth) — The gate intentionally does NOT call migrateLegacyAuthStore() before evaluating signedIn. Mirroring launch's resolveAuthEnv migration here is mutually exclusive with the prior round's fix: the sign-out refresh hits /observer/theaters, so an unconditional migrate would copy a just-removed legacy key back into the current store and resurrect a provider the user explicitly signed out of. The underlying cause is a known fleet-infra/auth SSoT limitation (migrate copies without deleting the legacy original, and deleteApiKey only clears the current store), which is already tracked as a separate follow-up rather than this PR. The gate is a best-effort current-store check: a legacy-only user is normalized as soon as /model-auth/state runs its migrate on Settings entry, and the 409 is a safe failure (no data loss / security impact). A proper read-only legacy lookup belongs in the dedicated fleet-infra/auth fix.

## Validation
- pnpm --filter @dotobokuri/fleet-console typecheck — pass
- pnpm --filter @dotobokuri/fleet-console test — pass (414)
- pnpm --filter @dotobokuri/fleet-console build — pass

sbluemin:
@codex The latest feedback has been addressed. Please re-review.

## Addressed
- P2 (operation-launch-menu.tsx:147) — Moved the FLEET_TERMINAL_CMD bypass into buildAgentCliLaunchMetadata so it covers both the launch menus and the server gate from one place. When the override is set, every Agent CLI is reported available + signedIn, so the client menus no longer disable rows the API would accept, and the server gate passes consistently. Only the boolean state is exposed (no command value).

## Validation
- pnpm --filter @dotobokuri/fleet-console typecheck — pass
- pnpm --filter @dotobokuri/fleet-console test — pass (414)
- pnpm --filter @dotobokuri/fleet-console build — pass

PR #114, opened 2026-06-24
feat(fleet-console): dock panel reorder and custom accent colors

sbluemin:
## Summary
- Dock taskbar panels can now be reordered by dragging a chip to a new position (or with Alt+Shift+Arrow while a chip is focused); the arrangement persists per Theater via a new `panelOrder` overriding the createdAt default.
- Each dock chip can be tinted with a custom accent color from a curated palette (`panelAccent`); the whole chip fill is tinted as a low-chroma "identity tint" while every system status channel (active brass, underway aurora/warn rings/glow, count badge, close coral, beacon) stays accent-immune. The accent picker is a body-portaled popover so it escapes the dock's `overflow:hidden` clip.
- Client-only change inside `runtime/fleet-console` plus a one-line doctrine reinforcement in the console `AGENTS.md` (user-accent owns fill only; no leak into signal channels).

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console test` (468 passed)
- [x] Real-browser console-e2e: drag lands at drop position + persists across refresh; Alt+Shift+Arrow keyboard reorder; accent applies on real click + persists; whole-chip tint reads naturally; active+accent coexistence; accent popover clickable (clip escaped); system signal channels unaffected
- [x] `node scripts/compile-changelog-fragments.mjs --check` (2 entries validated)

## Changelog
- [x] Includes a `.changelog.d/*.md` fragment (`dock-reorder-accent.md`, section: Added)


sbluemin:

## Addressed
- [P2] Keep shell dock metadata tab-scoped — canvas-store.ts `prunePanels`: shell ids are tab-scoped (sessionStorage), so pruning now always keeps `shell:*` ids in the shared `panelOrder`/`panelAccent` and only drops invalid Operation ids. A second tab can no longer delete the first tab's shell order/colors. A closed shell is still cleaned individually via `forgetPanelMetadata` on close. Added a regression test ("does not drop another tab's shell order/accent when pruning sees no shells").

## Validation
- `pnpm --filter @dotobokuri/fleet-console test` — 469 passed
- Real-browser console-e2e — accent perimeter ring, 16-color palette, carry-feel drag, active+accent coexistence verified

## Notes
- This PR also reworked the accent UX per Admiral direction: the accent is now a static colored ring on the chip's whole perimeter (the same `::before` edge band the in-progress orbit rides), not a fill tint or offset outline; palette expanded to 16 colors; drag now lifts and tracks the pointer for a carry feel.

sbluemin:
@codex All review feedback has been addressed and the branch is rebased onto the latest canary. Please re-review.

## Addressed
- [P2] Keep shell dock metadata tab-scoped — `prunePanels` always keeps `shell:*` ids and only drops invalid Operation ids; a closed shell is cleaned individually via `forgetPanelMetadata`.
- [P2] Preserve cross-tab shell order when reordering — `setPanelOrder` now merges hidden `shell:*` ids from the shared order instead of overwriting them with the visible-only list.
- [P2] Exclude the dragged chip from drop hit-testing — `dropIndexFromPoint` skips the source id, so a rightward drag from the left half of a chip lands correctly.

## Also in this revision
- Dock accent now persists in server durable state (`DurableOperation.accent`, exposed on the session DTO as a palette key) and is set through the existing `PATCH /terminal/sessions/:id` route as an independent partial update, so the chosen color survives a console restart (the loopback port, and thus the localStorage origin, changes on every start). The browser treats the server value as the source of truth with localStorage as an optimistic cache.
- Rebased onto canary, integrating the two-step close confirm (#112) with the reorder/accent surfaces on the same chip.

## Validation
- `pnpm --filter @dotobokuri/fleet-console typecheck` / `test` (473 passed) / `build` — all pass
- Server-level restart persistence verified end to end: PATCH accent -> state.json -> stop/start (new port) -> session DTO and the browser chip still show the accent on a fresh origin.

## Notes
- Operation order persistence across restart is a documented follow-up (same root cause, separate per-Theater route); reorder still persists per Theater across refresh.

sbluemin:
@codex Addressed the latest feedback. Please re-review.

## Addressed
- [P2] Preserve accent across SSE session updates — `readSessionInfo` in `client/src/sse.ts` now parses the `accent` field, so `session:updated` frames carry the server value instead of an `undefined` that the store would treat as a clear. Added a regression assertion in `sse.test.ts`.

## Validation
- `pnpm --filter @dotobokuri/fleet-console typecheck` / `test` (473 passed) / `build` — all pass

sbluemin:
@codex Addressed the latest feedback. Please re-review.

## Addressed
- [P2] Avoid re-uploading stale local accents — dropped the local->server upload migration in the Operations hydration effect (it was gated only by a remount-resetting ref, so returning after another tab cleared an accent resurrected the stale local value). The server durable accent is now the sole authority: hydration only aligns the local optimistic cache to the server value (set or clear), and accent reaches the server solely through the explicit chooseAccent PATCH.

## Validation
- `pnpm --filter @dotobokuri/fleet-console typecheck` / `test` (473 passed) / `build` — all pass

PR #143, opened 2026-06-30
fix(fleet-console): surface ALERTS notifications for non-active operations

sbluemin:
## Summary
- ALERTS now surfaces Awaiting/Complete notifications for every operation **except the one being actively viewed**, regardless of its Theater or minimized/maximized state. Previously the Operations canvas marked every operation as `visible` while mounted, so the visibility filter suppressed all notifications and the rail stayed empty.
- ALERTS "Open" reuses the same `focusOperation` + `navigate` path as the Cmd/Ctrl+K quick-search, which already switches to the operation's Theater and focuses the panel. Restoring notification visibility therefore also restores navigation to other-Theater panels (no separate move-logic change was needed).

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console test` (458 passed / 22 skipped)
- [x] Live console-e2e (notification surfacing + Open navigation across Theaters) recommended before/after merge

## Changelog
- [x] `.changelog.d/alerts-active-visibility.md` (Fixed)

### DIFF
--- a/.changelog.d/alerts-active-visibility.md
+++ b/.changelog.d/alerts-active-visibility.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- [fleet-console] ALERTS now surfaces Awaiting and Complete notifications for every operation except the one you are actively viewing, regardless of its Theater or minimized/maximized state; previously notifications were suppressed for all operations while the Operations canvas was open, so the rail stayed empty.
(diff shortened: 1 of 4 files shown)


sbluemin:

## Addressed
- **P2 (`notification-reduce.ts`) - Keep off-theater alerts visible.** `computeVisibleOperationIds` now treats the active operation as visible only when it actually exists in the current Theater. A `setActiveTheater` that leaves `activeOperationId` pointing at a previous Theater's (or a closed) operation no longer suppresses that operation's alert.
- **Field feedback - closed-panel alerts.** Added `filterByLiveOperations` so ALERTS drops notifications whose operation no longer exists (a panel closed after going awaiting/complete); this stops them from lingering and from producing a no-op when Open is clicked.

## Validation
- `pnpm --filter @dotobokuri/fleet-console typecheck`
- `pnpm --filter @dotobokuri/fleet-console test` - 461 passed / 22 skipped (added stale-active, closed-active, and live-filter cases)

## Notes
- No items declined or deferred.

PR #228, opened 2026-07-11
fix(fleet-console): recover packaged desktop launch and console procurement

sbluemin:
## Summary
The packaged desktop app never reached the console. Three defects, surfaced by adding real error logging:

- **Window never appeared** — the `GrantFileProtocolExtraPrivileges` fuse was disabled, revoking the renderer's privilege to load the entry HTML from inside `app.asar` over `file://` → `ERR_FILE_NOT_FOUND` before the window is shown → `app.exit(1)`. The shell only ever loads its own trusted, scriptless CSP-locked entry page (the console runs over http loopback), so the fuse is re-enabled. All other hardening fuses (RunAsNode/NodeOptions/NodeCliInspect off, ASAR integrity + OnlyLoadAppFromAsar on) stay.
- **Console runtime install failed** — npm lifecycle scripts (`node-pty` prebuild / node-gyp) resolve `node` from PATH via `sh -c "node ..."`, but a Finder/tray launch has no system node on PATH and npm does not add the running node's directory. `createConsoleInstallerEnvironment` now prepends the bundled node bin to PATH (POSIX + Windows).
- **Failures were silent** — startup and console-procurement errors (child-process `stderr`, exit code, and the full `cause` chain) are now written to the desktop log via `describeError`. Finder/tray launches send stderr nowhere, so this file log is both the dev-diagnosis and the published-app user-issue SSoT.

## Known follow-up (not in this PR — needs a decision)
Even with these fixes, the packaged app cannot finish booting against the **currently published** `@dotobokuri/fleet-console@1.23.0`, which predates `desktop-protocol.mjs` (tagged 2026-07-10; the file was added 2026-07-11). The desktop `verifyInstallation` requires `dist/desktop-protocol.mjs`, which no published console version ships yet. A console release > 1.23.0 (current source builds and includes it) is required for end-to-end boot. This is a release-sequencing item, tracked separately.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console-desktop typecheck`
- [x] `pnpm --filter @dotobokuri/fleet-console-desktop test` (108 passed; adds installer PATH-prepend POSIX/Windows + `describeError` coverage)
- [x] `pnpm desktop:package:dir` → fuses read back with `GrantFileProtocolExtraPrivileges` enabled; app launches and shows the entry window (previously crashed with `ERR_FILE_NOT_FOUND`)
- [x] Minimal-PATH launch reproduces the real install cause in the desktop log (`code=ENOENT ... desktop-protocol.mjs`), confirming npm install now succeeds and only the release gap remains

---
## Also in this PR: package rename
Renames the desktop workspace package `@dotobokuri/fleet-console-desktop` -> `@dotobokuri/fleet-desktop` and its directory `runtime/fleet-console-desktop` -> `runtime/fleet-desktop`. Pure internal rename (the package is not published to npm); the user-facing app product name `Fleet Console` and the appId `com.dotobokuri.fleet-console` are unchanged. All references updated (pnpm workspace, root scripts, CI workflows, fleet-cli desktop command, docs/AGENTS) and the lockfile regenerated. Verified: renamed package typecheck + 108 tests, dependent fleet-cli `desktop-command` + fleet-console suites, and `pnpm desktop:package:dir`.


sbluemin:

## Addressed
- **P2 — Preserve exact PATH on POSIX** (console-installer.ts): case-insensitive path-key matching is now win32-only; on POSIX only the exact `PATH` key is treated as the path variable, so a lowercase/mixed-case `path` is preserved as a distinct variable and the bundled node dir is never written to the wrong key.
- **P2 — Assert the file protocol fuse** (verify-packaged-app.mjs): `assertFuses` now also requires `GrantFileProtocolExtraPrivileges` to be enabled, so an artifact where the fuse was not applied fails `verify:package` instead of shipping the `ERR_FILE_NOT_FOUND` startup crash.

## Validation
- `pnpm --filter @dotobokuri/fleet-desktop typecheck` — pass
- `pnpm --filter @dotobokuri/fleet-desktop test` — 109 passed (adds a POSIX case-sensitive PATH regression test)
- `pnpm --filter @dotobokuri/fleet-desktop verify:package` — pass (fuse assertion satisfied by the packaged app)

## Notes
- This PR also renames the package `@dotobokuri/fleet-console-desktop` -> `@dotobokuri/fleet-desktop` (directory `runtime/fleet-console-desktop` -> `runtime/fleet-desktop`); the review comments re-anchored onto the new paths.

sbluemin:
@codex One more change was pushed after the previous approval — please re-review.

## Addressed (new commit)
- **Strip mac updater artifacts from release builds**: electron-builder emits `.blockmap` (and `latest*.yml`) for the mac dmg/zip targets, but the shell-only desktop does not use electron-updater and `verify:package` forbids updater artifacts, so `package:release` would fail post-package verification. Added an `afterAllArtifactBuild` hook that removes them from the output dir (nsis was already handled via `differentialPackage: false`).

## Validation
- `pnpm --filter @dotobokuri/fleet-desktop test` — 111 passed (adds strip-updater-artifacts unit tests)
- `pnpm --filter @dotobokuri/fleet-desktop package:unsigned` — builds the dmg/zip; the hook logs `stripped updater artifacts (shell-only): ...dmg.blockmap, ...zip.blockmap`
- `pnpm --filter @dotobokuri/fleet-desktop verify:package` — pass (no updater artifacts remain)

sbluemin:
@codex Rebased onto the latest `canary` and pushed a follow-up — please re-review.

## Changes since approval
- **Rebased onto `canary`**: integrates the merged `assertShellOnlyAsar` Windows ASAR path normalization (from #d424dabdf) with the fuse changes in `verify-packaged-app.mjs`; both are preserved (different functions). `verify:package` passes on the merged file.
- **Typecheck fix**: added a co-located `.d.mts` declaration for the `strip-updater-artifacts.mjs` build hook so the unit test's import is typed (TS7016 was slipping past vitest, which does not typecheck).

## Validation
- `pnpm --filter @dotobokuri/fleet-desktop typecheck` — pass
- `pnpm --filter @dotobokuri/fleet-desktop test` — 111 passed
- `pnpm --filter @dotobokuri/fleet-desktop verify:package` — pass

PR #279, opened 2026-07-17
feat(fleet-desktop): add managed SSH remote runtime

sbluemin:
## Summary
- **Connect to Runtime**을 2모드(Loopback address · Remote over SSH)로 확장합니다. Remote over SSH를 선택하면 Desktop이 SSH로 원격 머신에 Node와 Fleet Console을 자동 설치·구동하고, `ssh -L` 터널로 연결한 뒤 기존 loopback pairing으로 창을 넘깁니다. 페어링 입력은 JS-free 네이티브 modal, 실행·터널·창은 Desktop이 소유합니다.
- **원격 진행을 entry 부트스트랩 화면에 표시**합니다(로컬 부팅과 같은 boot-step). 실패 시 로컬 Console로 복귀하며 원인별 실패 dialog를 띄우고, 메뉴/트레이에 **Back to Local Runtime**(원격 연결 중일 때만 활성)을 추가합니다.
- 원격 플랫폼은 Linux x64/arm64와 macOS(darwin) x64/arm64를 지원합니다. Console 서버·`@fleet-console/desktop-protocol`·entry HTML/CSP는 무변경이며, loopback 게이트·token-free URL 등 보안 계약을 그대로 유지합니다. 신규 원격 오케스트레이션은 `src/runtime/remote/` 아래 Electron 비의존 순수 Node 모듈로, SSH argv 인젝션·경로 탈출을 고정 스크립트 레지스트리와 인자 스키마로 차단합니다.

## Test Plan
- [x] `pnpm --dir runtime/fleet-desktop typecheck && build && test` (227 tests green)
- [x] docker ubuntu(linux-arm64) live 게이트 8/8 체크포인트 통과
- [x] macmini(darwin-arm64) live 게이트 3회 연속 8/8 통과 (사용자 durable state 백업→복원, 해시 불변)
- [x] sentinel correctness 리뷰 반영(HIGH previousUrl 복구·MEDIUM switchTo single-flight)
- [ ] Changelog fragment: PR 번호 배정 후 `.changelog.d/pr-<number>.md` 추가 예정


sbluemin:

## Addressed
- P1 (orchestrator.ts) — Offline registry no longer skips lock inspection. An offline `registry.check` result now inspects the remote lock version-agnostically and reuses a live same-owner Desktop Console instead of provisioning/starting (which reclassified the live lock as same_owner and killed it). Foreign/CLI locks remain fail-closed; online version matching is unchanged.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green
- Live gate on a disposable macmini (darwin-arm64) exercised the same-owner reuse path

## Notes
- Also fixed in this PR (surfaced during live use): remote Agent CLI discovery. `start_console` now injects the remote login+interactive shell PATH (mirroring the local desktop's login-shell PATH handling), so the Console detects Claude/Codex on the remote host that were previously shown as "NOT INSTALLED"; verified on macmini that both resolve.

sbluemin:
@codex Both review comments have been addressed. Please re-review.

## Addressed
- P2 (orchestrator.ts / ssh.ts) — Stale remote lock is now cleared before reconnect. When initial inspection classifies the remote lock as `stale` (dead pid), the orchestrator runs a fixed `remove_console_lock` operation before provisioning, so a new `serve` can create its O_EXCL lock instead of timing out. Removal is limited to the `stale` branch only; absent/same-owner/foreign/CLI/conflict paths are untouched.
- P2 (ssh.ts) — SSH process handles now resolve on `close` (after stdio drains) instead of `exit`, so fast commands like `detect_platform` and `start_console` no longer intermittently return truncated stdout (which could misreport unsupported platform or an invalid pid).

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (235 tests)
- Live gate on a disposable macmini (darwin-arm64) with a pre-seeded stale lock: all 8 checkpoints passed — the stale lock was removed and the reconnect succeeded, where it previously required a manual delete

sbluemin:
@codex Both review comments have been addressed. Please re-review.

## Addressed
- P2 (ssh.ts / orchestrator.ts / lock.ts) — Stale lock removal is now pid-conditional. `inspectRemoteLock` retains the parsed lock; the orchestrator passes only that stale pid to `remove_console_lock`, which deletes the file only when its JSON pid still matches and the process is dead (`grep '"pid": <n>,' && ! kill -0`). If another process created a fresh/active lock in the meantime, the pid mismatch or live process leaves it untouched (fail-safe: it never deletes an active lock).
- P2 (tunnel.ts) — The default settle no longer returns immediately. It waits for either the SSH process to exit or a bounded 500ms window (`Promise.race`), so an `ExitOnForwardFailure` bind failure on an occupied local port is observed and rerolled instead of committing against a different local Console.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (238 tests)
- Lock JSON pid-line format (`"pid": <n>,`) confirmed against a real remote lock in earlier live runs; unit tests cover dead-match deletion vs live/mismatch retention and in-window bind-failure collision

sbluemin:
@codex All three comments have been addressed. Please re-review.

## Addressed
- P1 (ssh.ts) — Remote Node/npm now run under a sanitized environment. `start_console` and `install_console` `unset NODE_OPTIONS`, the Desktop control variables, and `ELECTRON_*` before executing node/npm, mirroring the local `sanitizeEnvironment`; the login-shell PATH is still hydrated first, and start_console sets only our explicit owner variables afterward. Verified on a real remote shell that injected `NODE_OPTIONS`/`FLEET_CONSOLE_DIR`/`ELECTRON_RUN_AS_NODE` are neutralized.
- P2 (ssh.ts) — `promote_runtime_path` now reconciles an interrupted promotion first: if `$2` is missing but `$2.old` exists, it restores `$2.old` to `$2` before proceeding, so an interrupted update no longer deletes the only working runtime.
- P2 (orchestrator.ts) — For a same-owner version mismatch, the replacement is fully provisioned before the live service is stopped; a provisioning failure leaves the previous runtime and its session intact.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (241 tests)
- Remote env sanitize verified live on macmini (darwin-arm64)

sbluemin:
@codex Addressed. Please re-review.

## Addressed
- P2 (ssh.ts) — `install_console` now strips inherited npm config before invoking npm: `for v in $(env | grep -iE '^npm_config_' | ...); do unset "$v"; done`, so a remote `npm_config_ignore_scripts=true`/`NPM_CONFIG_IGNORE_SCRIPTS=true` can no longer skip the node-pty lifecycle. Our explicit `npm_config_userconfig`/`globalconfig`/`registry` are applied as the exec prefix after the strip (mirrors the local installer stripping `npm_config_*`).

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green
- Verified live on macmini that injected `NPM_CONFIG_IGNORE_SCRIPTS` and `npm_config_registry` are neutralized by the strip

sbluemin:
@codex Addressed. Please re-review.

## Addressed
- P2 (runtime-pairing.ts) — A failed switch now preserves whatever session was active. The failure path branches on the previous origin: if it was the local runtime it restores local (unchanged); if it was a remote (A) it restores A's URL and window policy and keeps the committed A session handle instead of disposing it. Only the failed candidate (B) is rolled back, so trying to reach remote B while on remote A no longer tears down A.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (242 tests)
- Unit tests cover local->remote failure restore, remote A->remote B failure keeping A, and remote A->local switch

sbluemin:
@codex Addressed. Please re-review.

## Addressed
- P2 (runtime-pairing.ts) — Re-submitting the ssh target you are already connected to is now a no-op at the pairing layer: when `committedRemote.target` equals the parsed ssh target, switchTo returns early without calling connectRemote, opening a new tunnel, or switching the window, so the existing tunnel/port is never rerolled and the committed remote service is not stopped. Different targets and Back-to-Local are unaffected. This blocks the orchestrator reroll scenario at its root (the orchestrator has no knowledge of the committed session).

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green
- Unit tests cover same-target no-op (no connect/tunnel, committed session kept), different target switch, Back-to-Local, and first connect

sbluemin:
@codex Addressed. Please re-review.

## Addressed
- P3 (pairing-modal.ts) — The remembered SSH target now actually prefills the modal. After `loadFile`, the main process injects via `webContents.executeJavaScript` (the same one-way pattern entry-page uses under `script-src 'none'`) to select the Remote over SSH tab and set the host input to the saved target (with the `ssh:` prefix stripped). The injected value is escaped (JSON-based) against injection. The static HTML stays JS-free — no script, event handlers, or inline JS were added, and the private-scheme submit/cancel contract is unchanged.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (245 tests)
- Tests cover tab+host injection, quote/HTML-terminator escaping, and no injection when nothing is remembered; JS-free/CSP asset checks remain green

sbluemin:
@codex Both addressed. Please re-review.

## Addressed
- P2 (orchestrator.ts) — Each readiness pairing-identity fetch now carries a per-attempt `AbortSignal.timeout(min(3s, remaining deadline))`, combined with the external cancellation signal. A hung Console (TCP accepted, identity never completes) now aborts the attempt and the overall 20s deadline reports `remote_pairing_not_ready` instead of hanging in `verifying_pairing`.
- P2 (runtime-pairing.ts) — Same-target re-submit is now liveness-gated: it verifies `committedRemote.origin` (short timeout). If the tunnel is alive it stays a no-op (does not reroll/kill the live session); if it is dead (sleep/ssh exit/remote restart), it disposes the stale handle and reconnects, so the user recovers by selecting the same target.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (247 tests)
- Tests cover hung-identity attempt timeout -> deadline failure, live same-target skip, and dead same-target dispose->reconnect

sbluemin:
@codex Both addressed. Please re-review.

## Addressed
- P1 (pairing-modal.ts) — Removed all renderer JavaScript from the pairing modal, restoring the AGENTS.md "contains no JavaScript" boundary. The remembered SSH target is now prefilled by main-process HTML substitution: the template is read, the host is injected as an HTML-attribute-escaped `value=` and the Remote-over-SSH radio as `checked`, written to a 0600 temp file, loaded via `loadFile`, and the temp file is cleaned up on close. The renderer stays a pure JS-free static page; private-scheme submit/cancel and CSP are unchanged.
- P2 (tunnel.ts) — The `-L` forward now binds explicitly to loopback: `127.0.0.1:${port}:127.0.0.1:${port}`, so a client `GatewayPorts yes` cannot expose the token-free Console tunnel on LAN interfaces.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (247 tests)
- Tests: HTML-substitution prefill with escaping + temp-file cleanup + static fallback; JS-free/CSP asset checks; exact loopback `-L` argv snapshot

sbluemin:
@codex Both addressed. Please re-review.

## Addressed
- P2 (runtime-pairing.ts) — On SSH failure the local restore now reloads the exact previous route. It verifies the local origin, then loads `previousUrl` when it is same-origin and under `/console/`; otherwise it falls back to the base `/console/`. A failed remote connect from `/console/settings` etc. no longer drops the user to the default route.
- P3 (pairing-modal.ts) — The remembered-prefill temp page is now created in an `mkdtemp` directory with `pairing.css` copied alongside (mode 0600), so the relative stylesheet resolves and the tabs/panels are styled; the whole temp directory is cleaned up on close. Static (no-remembered) path unchanged.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green (249 tests)
- Tests: local-route preservation + foreign-origin fallback; temp index.html + css copy + permissions + directory cleanup

sbluemin:
@codex Addressed. Please re-review.

## Addressed
- P2 (ssh.ts) — Remote uploads no longer crash the main process on a broken pipe. An `error` handler is attached to `child.stdin` (and removed once the process exits), so an EPIPE from an early SSH exit during `upload_file` (dropped connection, permission, full remote disk) is absorbed instead of surfacing as an unhandled stream error. The failure is reported normally: the child exits non-zero and `collect` returns `ssh_failed`, which the orchestrator rolls back.

## Validation
- `pnpm --dir runtime/fleet-desktop typecheck && build && test` — green
- Tests: EPIPE/stdin write error during upload -> no unhandled crash, ssh_failed via non-zero exit, handler attached

PR #416, opened 2026-07-26
feat(console): surface console link loss across the chrome

sbluemin:
## Summary

- 콘솔 서버가 죽어도 전역 배너·토스트·모달이 하나도 뜨지 않았습니다. 실측하면 저장소 패널이 브랜치 8→0, 태그 124→0, 기록 200행→"기록 없음"으로 **오류 표시 없이 빈 값으로 리셋**되어, 빈 저장소와 죽은 서버가 화면에서 구별되지 않았습니다. `connectionError` 상태는 Toast 배선까지 있었지만 설정하는 호출부가 0건인 데드코드였습니다.
- 링크 상태를 세 층으로 알립니다: command band 상태 칩, 마지막 갱신 시각을 고정하는 배너, 값이 멈춘 레일 패널 위의 stale 덮개. 각 층에 재연결 컨트롤이 있습니다. **연결이 정상이면 셋 다 렌더되지 않아** 평상시 크롬은 그대로입니다.
- 활성 EventSource와 그 generation을 한 곳에서 소유합니다. 이전 연결을 닫지 않던 재연결 경로가 여러 살아있는 연결을 만들어 상태를 경합시키던 문제를 없앴습니다.
- 덮개가 떠 있는 동안 패널 콘텐츠를 `inert`로 만들고, 포커스를 재연결 컨트롤로 넘겼다가 링크가 살아나면 원래 요소로 되돌립니다.

## Test Plan

- [x] 집중 테스트 5 files, 59 tests 통과 (`connection-state`, `connection-ui-contract`, `operations-sse`, `right-rail`, `instrument-design-contract`)
- [x] `pnpm typecheck` — `tests/server.test.ts:2623,2648` TS2741 2건만 남음(canary 선존, 이 변경과 무관)
- [x] 격리 실브라우저 — 연결 정상: 칩·배너·덮개 전부 미렌더
- [x] 격리 실브라우저 — 서버 SIGKILL: 3층 전부 표시, 배너 `role=status`/`aria-live=polite`, 최초 단절 시각이 재시도 중에도 유지
- [x] 격리 실브라우저 — 재연결 연타 4회: 열린 EventSource **항상 0개**(수정 전에는 `closed:false` 2개 잔존)
- [x] 격리 실브라우저 — 단절 시 패널 콘텐츠 `inert=true`, 포커스가 덮개 재연결 버튼으로 이동
- [ ] 링크 복구 시 포커스 복원 — 단위 테스트와 캐리어 실측은 통과했으나, 호스트가 독립 재현하지 못했습니다(격리 서버 재기동 시 포트가 바뀌어 같은 페이지로 복구를 유도할 수 없고, loopback SSE는 브라우저 오프라인 토글로 끊기지 않음)
- [x] Changelog fragment — `.changelog.d/pr-416.md` (grouped 문법, EN/KO 병기, `--check` 통과). 덮개 포커스 항목은 이 PR이 도입한 미출시 동작이라 Release Baseline 규칙에 따라 독립 Fixed가 아닌 Added로 접어 넣었습니다.

### DIFF
--- a/.changelog.d/pr-416.md
+++ b/.changelog.d/pr-416.md
@@ -1,0 +1,6 @@
+### fleet-console
+#### Added
+- [fleet-console] The console now reports a lost server link on the command band, in a banner pinning the time values stopped updating, and over rail panels whose values are frozen, each offering to reconnect.
+  ko: 콘솔이 서버 연결이 끊긴 사실을 command band와, 값이 멈춘 시각을 고정한 배너와, 갱신이 멈춘 레일 패널 위에서 알리고 각각 다시 연결할 수 있게 합니다.
+- [fleet-console] While that overlay covers a rail panel, its contents stay out of the keyboard tab order and focus moves to the reconnect control, returning once the link is live.
+  ko: 그 덮개가 레일 패널을 가리는 동안에는 패널 내용이 키보드 탐색에서 빠지고 포커스가 다시 연결 컨트롤로 옮겨가며, 연결이 살아나면 원래 자리로 돌아옵니다.
--- a/runtime/fleet-console/core/client/src/app.tsx
+++ b/runtime/fleet-console/core/client/src/app.tsx
@@ -20,7 +20,7 @@ import { GlobalSettings } from "./pages/global-settings.js";
 import { Operations } from "./pages/operations.js";
 import { BUILT_IN_RAIL_PANELS } from "./rail/built-in-panels.js";
 import { toggleRailChrome } from "./rail/rail-store.js";
-import { refreshObserverStatus } from "./operations-sse.js";
+import { reconnectOperationsSseNow, refreshObserverStatus } from "./operations-sse.js";
 import { closeKeyboardShortcuts, hydrateGroups, hydrateInitialOperations, hydrateOperations, hydrateTheaterBootstrap, hydrateTheaters, resolveOnboardingOnBootstrap, setOperationsViewActive, setState, toggleOperationSearch } from "./store.js";
 import { abortReleaseNotesFetch, requestReleaseNotes } from "./release-notes-fetch.js";
 import { getSideBarState, setSideBarCollapsed, subscribeOperationActivityTracking } from "./sidebar/operations-side-bar-store.js";
@@ -48,6 +48,9 @@ export function App() {
   const t = useT();
   const consoleLocale = useConsoleLocale();
   const releaseNotesLocale = resolveReleaseNotesLocale(globalSettings.state?.language ?? "auto");
+  const connectionLostTime = state.connectionLostAt === null
+    ? ""
+    : new Date(state.connectionLostAt).toLocaleTimeString(consoleLocale);
 
   useEffect(() => {
     document.documentElement.lang = consoleLocale;
@@ -216,6 +219,12 @@ export function App() {
   return (
     <div className="console-shell">
       <CommandBand operationsViewVisible={operationsViewVisible} />
+      {state.connection === "offline" ? (
+        <div className="console-link-banner" role="status" aria-live="polite">
+          <span>{t("chrome.link.offline")}. {t("chrome.link.bannerDetail", { time: connectionLostTime })}</span>
+          <button type="button" onClick={reconnectOperationsSseNow}>{t("chrome.link.reconnect")}</button>
+        </div>
+      ) : null}
       <main className="console-route-content">
         <Routes>
           <Route path="/" element={<Navigate to="/operations" replace />} />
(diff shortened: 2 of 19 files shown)


sbluemin:

## Addressed
- P2 `operations-sse.ts:92` — Hydrate operations during manual reconnection. The retry timer was the only caller of `fetchOperations()`, and a manual reconnect cancelled it, so changes made during the outage were never recovered. Folded the snapshot into `source.onopen` under the existing `isCurrentSource()` guard, so every reconnect path (manual and automatic) resyncs exactly once on connect; removed the now-redundant pre-fetch from the retry timer. Backoff (1s→30s) and the 1s reset on open are unchanged.
- P2 `right-rail.tsx:369` — Preserve focus when it originated outside the stale panel. Restoration now returns early when the veil never captured focus, so a disconnect that happens while the user is in the sidebar or command band no longer pulls focus into the rail on recovery. The capture-side condition (only when focus was inside the panel content) was already correct and is unchanged; the panel-body fallback is kept for the case where the captured element disappeared, since focus would otherwise fall to `body`.

## Validation
- `vitest run` on connection-state / connection-ui-contract / operations-sse / right-rail / instrument-design-contract — 5 files, 61 tests passed
- `pnpm build` — passed
- `pnpm typecheck` — only the two preexisting `tests/server.test.ts:2623,2648` TS2741 failures remain (present on canary, unrelated to this PR)
- Isolated real browser: after SIGKILL and a manual reconnect, the snapshot included the operation created during the outage; exactly one active EventSource; zero page errors

## Notes
- Both findings reproduced before the fix; neither was declined.
- Not independently verified: the rendered list after a restart, because the isolated server binds a new port and the original page cannot be steered back to it. Covered by a regression test that fails if the fix is reverted.

sbluemin:

## Addressed
- P2 `operations-sse.ts:60` — Reconcile SSE frames received during snapshot hydration. Rather than buffering or merging frames, this rolls back the change that created the race: the previous pass moved the resync into `onopen`, which meant a snapshot could land *after* the live frame carrying a newer node. The original ordering is restored — `fetchOperations` → `hydrateOperations` → `connectOperationsSse` — so hydration always completes before the stream is open, and the manual reconnect path now follows that same order instead of connecting directly. Each path captures its own generation, so a late response neither hydrates nor connects once a newer attempt has started, and a failed fetch still proceeds to connect. The ordering contract is now stated in a comment so it is not folded back into `onopen`.
- P2 `right-rail.tsx:370` — Preserve focus after the user leaves the stale veil. Restoration now runs only while the focus is still the veil's own — the active element is inside the veil, or `body` because the control it held was removed. If the user moved to the command band, sidebar, or a rail tab during the outage, the captured target is cleared without moving focus.

## Validation
- `vitest run` on connection-state / connection-ui-contract / operations-sse / right-rail / instrument-design-contract — 5 files, 62 tests passed
- `pnpm build` — passed
- `pnpm typecheck` — only the two preexisting `tests/server.test.ts:2623,2648` TS2741 failures remain (present on canary, unrelated to this PR)
- Regression tests cover the ordering contract (hydration completes before `EventSource` is constructed; a stale-generation response does neither) and all four focus cases: restore inside the panel, fall back to the panel body when the captured element is gone, leave focus alone when the user moved away, and leave it alone when nothing was captured.

## Notes
- Neither finding was declined; both reproduced.
- The buffering / `ts.updatedAt` merge suggestion was not adopted. The race existed only because of the previous pass's fix, so the narrower correction is to restore the ordering that never had the race rather than to add reconciliation on top of it.

sbluemin:
@codex Thanks — this one is deferred rather than fixed, with reasoning below.

## Deferred
- P2 `operations-sse.ts:102` — Close the gap between snapshot and stream subscription. The gap is real: a mutation landing between the snapshot and the `EventSource` construction is not replayed. Three reasons it does not belong to this PR:
  1. **It predates this PR.** `origin/canary` already reconnects as `fetchOperations().then(hydrateOperations).finally(connectOperationsSse)`, so the unsubscribed interval exists on the base branch. This PR does not introduce or widen it — the manual path now follows the same ordering the automatic path always had.
  2. **The suggested alternatives are out of scope or regressive.** A replayable snapshot cursor is a server protocol change (`core/host/`), which this PR explicitly excludes; it is a change to how the operations stream is defined, not to how a lost link is surfaced. "Connect first and reconcile buffered frames" is exactly the arrangement the previous pass flagged: it lets a pre-change snapshot overwrite the newer node it raced. Swapping between the two orderings trades one window for the other — only a replay cursor removes both, and that is its own piece of work.
  3. **This PR's scope is reporting a lost link**, not making the operations stream gap-free. Everything it adds is presentational plus the connection ownership needed to make it correct.

Recorded as a follow-up: give the operations stream a replay cursor so a reconnect can resume from a known position instead of racing a snapshot.

## Status
Three review passes have run; this pass produced no finding that meets the fix bar, so the review loop stops here and the PR proceeds to merge after the final context audit. The two earlier passes were both fixed, and their reasoning is in the comments above.

PR #448, opened 2026-07-31
feat(fleet-plugins): add a daily cost trend chart to the Ledger panel

sbluemin:
## Summary

- Ledger 패널의 "CLI별 · 기기 전체" 섹션 상단에 일별 비용 추이 막대 차트를 추가합니다. 실측된 문제는 "이번 달 $12,891.72" 같은 총액만 보이고 **언제 얼마나 튀었는지**가 UI에서 완전히 소거된다는 점이었습니다(7월 10~11일 이틀에 $2,641.83 소비가 현재 UI에서는 보이지 않음).
- 서버 요약 DTO에 `daily` 필드(로컬 날짜별 비용)를 추가했습니다. 기존 파서가 이미 추출하던 `lastActive`를 버킷팅할 뿐이라 새 의존성이나 tokscale 업그레이드가 없습니다. `clients`와 같은 기기 전체 스코프이므로 theaterId로 필터링하지 않습니다.
- 배치 근거: 히어로는 Operation 귀속 스코프이고 시계열은 기기 전체 스코프라, 두 스코프를 같은 카드에 두면 사용자가 이미 혼동하는 지점을 재생산합니다. 그래서 기기 전체 섹션 상단에 둡니다.
- 창이 2일 미만이면(오늘 창) 차트는 렌더되지 않습니다.
- QA 지적 반영: 4자리 연도로 표현 불가한 타임스탬프는 파서에서 degraded 처리(패널 크래시 방지), 막대는 `role="img"` 컨테이너 안의 무동작 button 대신 `role="group"` + `span[role="img"][tabIndex=0]`, 일 평균은 오버플로 없이 계산.

## Test Plan

- [x] `pnpm --filter @fleet-plugins/ledger typecheck` — 통과
- [x] `pnpm --filter @fleet-plugins/ledger test` — 81 passed (기준선 69)
- [x] `pnpm --filter @fleet-plugins/ledger build` — 통과
- [x] `pnpm vitest run tests/instrument-design-contract.test.ts` (fleet-console) — 27 passed
- [x] 격리 콘솔 실브라우저 실측: 7일 창 막대 7개 / 이번 달 창 막대 31개(축 7월 1일–31일, 최고 7월 12일 $1,698.67 · 일 평균 $416.33) / 오늘 창 차트 은닉 / 한글 로케일 라벨·aria 라벨 확인
- [x] `TZ=Pacific/Kiritimati` 등 대체 타임존에서 summary 테스트 통과

## Changelog

- [x] `.changelog.d/pr-448.md` — 제품별/섹션별 문법, 영문+한글 요약 쌍, `node scripts/compile-changelog-fragments.mjs --check` 검증 완료


sbluemin:

## Addressed

- **P1 — session cost assigned to the last-active day** (`server/summary.ts`): fixed as a labeling change, not an attribution change. Measured on real data: 98 of 4,240 sessions cross local midnight and carry **26.7% of total cost** ($3,446.96 of $12,919.48), so the concern is material. However, tokscale exposes only per-session aggregates (`created_at` / `last_active` + totals) and no per-day usage, so a true split does not exist upstream; splitting proportionally by wall-clock would invent spend for hours a session sat idle. The panel now states the rule under the trend title, reusing the existing `.ledger-clients-description` grammar: "Each session's cost counts on the day it was last active, so a session spanning midnight lands entirely on the later day."
- **P1 — zero-cost days omitted** (`server/summary.ts`): interior gaps between the first and last observed day are now filled with zero-cost points, so bar spacing is proportional to real time and the daily average divides by charted days rather than active days. Day iteration uses local-date arithmetic (DST-safe), and the series is capped at the most recent 366 days because the parser admits any four-digit-year timestamp and one stale record could otherwise render thousands of nodes.
- **P1 — idle bars on the brass channel** (`client/ledger.css`): correct call, and the doctrine is explicit that brass is location/focus/hover only. Idle bars are now `var(--ink-fog)`; brass is retained for hover/focus fill and the focus outline, which is brass's own channel.

## Validation

- `pnpm --filter @fleet-plugins/ledger typecheck` — pass
- `pnpm --filter @fleet-plugins/ledger test` — 83 passed (was 81)
- `pnpm --filter @fleet-plugins/ledger build` — pass
- `pnpm vitest run tests/instrument-design-contract.test.ts` (fleet-console) — 27 passed
- `TZ=Pacific/Kiritimati` and `TZ=America/Los_Angeles` summary suites — 21 passed each
- Isolated Console, real browser: month window renders 31 bars with the attribution line, bar fill resolves to `oklch(0.52 0.012 245)` (neutral ink), hover/focus still brass

## Notes

- **Declined — "populate every calendar date in the requested window."** The plugin cannot know the window's true boundaries: a `--month` report run during July returns a `2026-06-30` bucket, so tokscale's windows are rolling, not calendar-aligned. Filling to an assumed calendar edge would either drop a day the source included or invent days it never queried. Filling interior gaps is provable from the returned data alone, so that is what was implemented.
- **Declined — "a window with only one active day suppresses the chart entirely."** The `daily.length >= 2` rule is intended: a single point is not a trend, and the Today window is meant to render no chart. Interior filling does not change this, and it should not.
- **Deferred — a dedicated data-visualization color channel.** The design doctrine enumerates signal, brass, and identity channels only, with no channel for passive data marks; neutral ink is the conservative fit today. Introducing a data-viz channel is a doctrine amendment and belongs in its own change, not here.

sbluemin:

## Addressed

- **P2 — edge tooltips clipped by the rail** (`client/ledger.css`, `client/rail-panel.tsx`): confirmed, and the measurement showed a wider blast radius than the comment described. In an isolated Console at the 392px rail, `.ledger-root` computes `overflow-x: auto` (because `overflow-y: auto` pushes the other axis off `visible`), its content box spans x=845..1236, bars are 9.58px, and tooltips are 92-104px — so **7 of 31 bars clipped**, not just the outermost two: indices 0/1/2/3 by 29/24/12/1px on the left and 28/29/30 by 2/17/25px on the right.

  Because of that, a `:first-child`/`:last-child` clamp would have been an incomplete fix. Instead each bar carries `--ledger-bar-pos = index / (length - 1)`, the bars row becomes the tooltip containing block, and the tooltip resolves as `left: calc(var(--ledger-bar-pos) * 100%)` with `transform: translateX(calc(var(--ledger-bar-pos) * -100%))`. The first tooltip's left edge aligns to the chart's left edge, the last one's right edge to the right edge, and everything between tracks its bar — with no bar-count, label-width, or locale assumptions and no magic numbers.

## Validation

- `pnpm --filter @fleet-plugins/ledger typecheck` — pass
- `pnpm --filter @fleet-plugins/ledger test` — 84 passed (was 83)
- `pnpm --filter @fleet-plugins/ledger build` — pass
- `pnpm vitest run tests/instrument-design-contract.test.ts` (fleet-console) — 27 passed
- Isolated Console, real browser, re-measured after the fix: month window (31 bars) **0 clipped tooltips**, 7-day window (7 bars) **0 clipped**. Middle-bar tooltip center lands at x=1035.5 against a bar center of x=1036, so the proportional anchoring stays visually centered away from the edges.

## Notes

- Not changed: the tooltip remains a per-bar reveal on hover/focus. Replacing it with a shared inspector line would sidestep edge positioning entirely, but that changes an approved interaction rather than fixing the reported defect, so it stays out of this PR.

sbluemin:
@codex Addressed — and correcting my previous reply.

## Correction

In the last round I declined "fill the complete selected window" on the grounds that tokscale's windows are rolling rather than calendar-aligned, citing a `--month` report that appeared to return a `2026-06-30` bucket. **That evidence was wrong.** It came from a throwaway measurement script of mine that derived day keys with `toISOString()` (UTC) instead of local time; in KST a local July 1 early-morning session maps to June 30 UTC. Re-measured in local time on 2026-07-31, the real ranges are:

| window | tokscale last-active range | derived from `scope.window` + `generatedAtMs` |
|---|---|---|
| `today` | 2026-07-31 .. 2026-07-31 | same |
| `week` | 2026-07-25 .. 2026-07-31 | `today-6 .. today` — same |
| `month` | 2026-07-01 .. 2026-07-31 | first of local month .. today — same |

So the derivation you proposed does hold, and the decline is withdrawn.

## Addressed

- **P1 — fill the complete selected window** (`server/summary.ts`): the window's local day range is now derived from `scope.window` + `generatedAtMs` and **unioned** with the observed range before filling, so leading and trailing quiet days become zero-cost points and the axis and daily average describe the whole window.

  The union is deliberate rather than using the derived boundary alone. On 2026-07-31 "calendar month" and "last 31 days" produce an identical range, so this data cannot pin which one tokscale implements. Taking `min(derivedStart, firstObserved) .. max(derivedEnd, lastObserved)` is correct under either: a wider upstream range is preserved instead of clipped, and a quiet start of window still extends the axis.

  Preserved deliberately: an empty result still yields `daily: []` rather than an all-zero chart, because the panel already has its own empty state. The `daily.length >= 2` render guard is unchanged — with window filling a week or month holding one active day now produces multiple points and renders naturally, while the `today` window still yields a single point and stays hidden.

## Validation

- `pnpm --filter @fleet-plugins/ledger typecheck` — pass
- `pnpm --filter @fleet-plugins/ledger test` — 87 passed (was 84)
- `pnpm --filter @fleet-plugins/ledger build` — pass
- `pnpm vitest run tests/instrument-design-contract.test.ts` (fleet-console) — 27 passed
- Summary suite under `TZ=Pacific/Kiritimati`, `TZ=America/Los_Angeles`, and `TZ=Australia/Lord_Howe` (a 30-minute offset zone) — 24 passed each
- Isolated Console, live plugin API: `today` 1 point (2026-07-31), `week` 7 points (2026-07-25..31), `month` 31 points (2026-07-01..31) — the served ranges now match the derived windows exactly

One honest limitation: this machine's real data has activity on every day of the current window, so the **zero-fill itself is proven by unit tests rather than by live observation**; what the live run confirms is the window range.

PR #465, opened 2026-08-02
feat(fleet-admiral): let the host assign a model per workflow stage

sbluemin:
## Summary

- Gateway 세션의 호스트가 워크플로 Phase마다 모델과 추론 강도를 배치할 수 있도록, 배치 가능한 로스터를 `gateway_models` MCP 도구 하나로 공급합니다. 노출 선별은 세션 실행 중에도 바뀌므로 호출 시점에 해석합니다.
- 응답은 세 계층을 섞지 않고 그대로 전달합니다 — 카탈로그가 진술하는 제약(core-ai-gateway), Fleet이 실측한 역할 적합도(fleet-admiral), 조회 시점의 프로바이더 잔여량(quota 플러그인). 도구는 이것을 판정으로 합치지 않습니다.
- Kimi가 사용량 패널의 네 번째 프로바이더로 합류하고, Cursor의 Auto/API 풀이 창 단위로 분리됩니다. 합산 수치는 특정 모델이 끌어 쓰는 풀이 고갈됐는지 답하지 못합니다.
- Delegation Policy에 배치 규칙을 명시하고(상속이 기본, 고정에 입증 책임), 절차는 온디맨드 `model-loadout` 스킬로 분리했습니다.

## Test Plan

- [x] `pnpm build` (저장소 전체)
- [x] `pnpm --filter @dotobokuri/core-ai-gateway test` — 171 passed
- [x] `pnpm --filter @dotobokuri/fleet-admiral test` — 96 passed / 1 skipped
- [x] `pnpm --filter @fleet-plugins/quota test` — 55 passed
- [x] `pnpm --filter @fleet-plugins/terminal test` — 526 passed
- [x] `pnpm check:core-boundary`
- [x] `GET /coding/v1/usages` 실측 응답으로 Kimi 파서 픽스처 작성 (2026-08-02)
- [x] Changelog: `.changelog.d/pr-465.md` — 제품/섹션 그룹 문법, 영문·한글 인접 요약, `node scripts/compile-changelog-fragments.mjs --check` 통과 (35 entries)

## Notes

- Kimi 사용량 엔드포인트는 공식 문서에 없습니다. 실패 시 해당 프로바이더만 `error`로 격리되며 다른 프로바이더에는 영향이 없습니다.
- `/v1/messages`는 노출 선별을 적용하지 않는 기존 동작을 유지합니다. 이번 변경은 MCP 응답 범위만 선별에 맞춥니다.
- `roleFit` 선언은 의도적으로 희소합니다. 실측이 7개 역할축 중 2개에서만 모델을 변별했고, 미측정 축은 `null`로 보고되어 상속으로 귀결됩니다.


sbluemin:
@codex The review feedback has been addressed, and the PR scope was extended by an explicit request from the repository owner. Please re-review.

## Addressed

- **Always include the Claude baseline when quota reads fail** — `packages/fleet-admiral/src/ai-gateway/model-loadout.ts`. Confirmed: `claude` serves no gateway model, so it entered the provider list only through the quota snapshot, and an unreadable snapshot removed it entirely. The baseline is now seeded unconditionally and reports `unsupported` like any other unreadable provider, so a host can distinguish "could not read" from "no such budget". Regression test: `keeps the parent baseline even when no allowance could be read at all`.
- **Give Kimi its own no-subscription message** — `runtime/fleet-plugins/quota/client/rail-panel.tsx`. Confirmed: only the Cursor and Kimi parsers reach `no_subscription`, and the shared string reads "No active Cursor subscription." Status copy is now selected per provider via `NO_SUBSCRIPTION_KEY`, matching the existing signed-out and expired maps. The Kimi wording states what the response showed ("Kimi returned no usage windows for this key.") rather than asserting a subscription verdict the payload cannot distinguish. Regression test: `names its own provider in every provider-specific message`.

## Scope extension (owner-directed)

Four on-demand operation skills were added for gateway sessions: `architecture-review`, `implementation-run`, `quality-review`, `codebase-research`. They carry the roles that carrier personas covered under classic doctrine, which gateway doctrine cannot render. Each skill owns only the shape of a run (stage skeleton, per-stage return contract, stopping rule) and names no model, leaving assignment to `model-loadout`. Classic sessions render none of them.

## Validation

- `pnpm build` (repository-wide) — pass
- `pnpm check:core-boundary` — pass
- `@dotobokuri/core-ai-gateway` — 171 passed
- `@dotobokuri/fleet-admiral` — 97 passed / 1 skipped
- `@fleet-plugins/quota` — 56 passed
- `@fleet-plugins/terminal` — 526 passed
- `node scripts/compile-changelog-fragments.mjs --check` — 36 entries validated

## Notes

- The gateway prompt has a size budget enforced by test. The Delegation Policy clause is deliberately compressed to policy only; the procedure and the measured evidence live in the on-demand skill so the always-on prompt does not carry them.
- `roleFit` declarations remain sparse on purpose. Measurement separated the models on two axes out of seven, so an unmeasured axis reports `null` and resolves to inherit rather than offering a reason to pin.

sbluemin:
@codex Both findings from the second pass are fixed. Please re-review.

## Addressed

- **Derive the inherited allowance from the session default** — `packages/fleet-admiral/src/ai-gateway/model-loadout.ts`. Confirmed at `runtime/fleet-plugins/terminal/server/agent-api/launch.ts:273`: a configured session default model is seeded as `ANTHROPIC_MODEL`, so unpinned work spends that model's provider allowance. The baseline is now marked explicitly with `isInheritedBaseline` and follows the session default when one exists, defaulting to the parent subscription otherwise. The parent stays in the roster either way, because switching the model back mid-session spends it again. The tool's usage guideline no longer asserts that the parent is always the baseline. Regression test: `moves the inherited baseline onto the session default model's provider`.
- **Allow the quota summary to finish its provider timeouts** — `runtime/fleet-plugins/terminal/server/agent-api/gateway-loadout.ts`. Confirmed at `runtime/fleet-plugins/quota/server/service.ts:101`: the summary resolves all four providers with `Promise.all`, each request tolerating 10s (`server/providers.ts`) and codex/cursor issuing two sequential calls, so a cold read's worst case is ~20s. The client timeout is now 25s, above that worst case, so providers that already answered are no longer discarded. Regression test: `waits longer than the summary's own worst-case provider wait`.

## Validation

- `pnpm build` (repository-wide) — pass
- `@dotobokuri/fleet-admiral` — 98 passed / 1 skipped
- `@fleet-plugins/terminal` — 527 passed

## Notes

- No changelog entry was added for these two. Both correct behavior introduced by this same PR, which is still unreleased, so folding them into the existing `pr-465.md` entries is the repository's release-baseline rule rather than emitting a standalone Fixed line.

sbluemin:
The third-pass finding is fixed, with a different remedy than the one suggested. Recording the reasoning, then merging.

## Finding: preserve the launched model when refreshing the live roster

The defect is real and reproduced: `readSelection` runs on every tool call while `ANTHROPIC_MODEL` is written once at launch (`agent-api/launch.ts:273-275`), so changing the Settings default mid-session made the roster name a provider the running session never switched to.

The suggested remedy — retaining the launch-time inherited model per session — is not available in this structure. `extraAgentTools` is registered once per runtime during plugin initialization (`server/agent.ts:113`), while launch runs per session through `registerLaunchResolver`. One tool spec serves every session, so there is no per-session place to hold a launch-time value. Threading session identity into the tool would add a new coupling to obtain a fact the host already has.

The deeper problem is that the roster was asserting something it cannot observe. It named an inherited baseline twice and was wrong both times — first by hard-coding the parent subscription, then by tracking the editable setting. Rather than widen the second attempt, it is removed: every provider's allowance is reported plainly, the parent subscription included, and the host matches its own running model against them. Claude Code knows what it is running; the roster does not need to guess.

Both earlier findings stay fixed by this: the parent allowance is still always listed, and nothing claims it is universally the baseline.

The usage guidance now states the limit explicitly, and records that `isSessionDefault` reflects the current setting rather than what a running session started with.

## Validation

- `pnpm build` (repository-wide) — pass
- `@dotobokuri/fleet-admiral` — 98 passed / 1 skipped
- Regression test: `does not claim which allowance an unpinned stage spends`

Merging at the repository owner's direction; no further review pass is being awaited.

PR #492, opened 2026-08-03
fix(gateway): strip format hints from OpenAI strict tool schemas

sbluemin:
## Summary
- codex 경로에서 Claude Code의 `WebFetch` 도구가 `url: { format: "uri" }`를 갖고 있어, strict 재작성 후 OpenAI가 요청 전체를 400(`'uri' is not a valid format`)으로 거부하는 결함을 수리합니다.
- OpenAI strict 모드는 `format` 키워드를 닫힌 값 집합에 대해 검증하므로, `format`을 `$schema`/`default`와 같은 "strictSchema가 와이어 도달 전 드롭하는 키워드" 그룹으로 이동했습니다. 자문적 힌트만 사라질 뿐 생략 보장(strict 재작성 ↔ null 스트립) 메커니즘은 불변입니다.
- 같은 메커니즘의 잔여 구멍인 `$defs` 미재귀를 봉합했습니다: `strictCompatible`은 `$defs` 하위 스키마를 허용 판정하면서 `strictSchema`는 정화하지 않아, `$defs` 안의 `format`/`default`/미재작성 객체가 동일 계열 400을 유발할 수 있었습니다.

## Test Plan
- [x] 관측된 WebFetch 거부를 재현하는 와이어 바디 회귀 테스트 추가 (`format: "uri"`가 strict 재작성 후 와이어에 남지 않음)
- [x] `$defs` 하위 스키마가 인라인 하위 스키마와 동일한 strict 정화를 받는지 회귀 테스트 추가
- [x] `packages/core-ai-gateway` vitest 219/219 통과
- [x] `tsc --noEmit` + `pnpm build` 통과

## Changelog
- [x] `.changelog.d/pr-492.md` — 문법·이중언어 요약·`compile-changelog-fragments.mjs --check` 검증 완료

### DIFF
--- a/.changelog.d/pr-492.md
+++ b/.changelog.d/pr-492.md
@@ -1,0 +1,4 @@
+### fleet-core
+#### Fixed
+- [core-ai-gateway] Strip JSON Schema `format` hints (such as `format: "uri"` on WebFetch's `url`) from OpenAI strict tool schemas, and apply the same strict cleanup to `$defs` subschemas, so a format value outside OpenAI's accepted set no longer fails the whole request with a 400 schema error on the Codex path.
+  ko: OpenAI strict 도구 스키마에서 JSON Schema `format` 힌트(WebFetch `url`의 `format: "uri"` 등)를 제거하고 `$defs` 하위 스키마에도 동일한 strict 정화를 적용하여, OpenAI가 허용하지 않는 format 값이 Codex 경로에서 요청 전체를 400 스키마 오류로 실패시키지 않도록 수정했습니다.
--- a/packages/core-ai-gateway/src/openai-responses-adapter.ts
+++ b/packages/core-ai-gateway/src/openai-responses-adapter.ts
@@ -638,7 +638,6 @@ const STRICT_ALLOWED_KEYWORDS = new Set([
   "$ref",
   "description",
   "title",
-  "format",
   "pattern",
   "minimum",
   "maximum",
@@ -652,6 +651,7 @@ const STRICT_ALLOWED_KEYWORDS = new Set([
   // Dropped by `strictSchema` before the schema reaches the wire.
   "$schema",
   "default",
+  "format",
 ]);
 
 /** Whether every subschema stays inside the subset strict mode accepts. */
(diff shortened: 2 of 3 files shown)


sbluemin:

## Addressed
- P1 (array-blind null strip) — packages/core-ai-gateway/src/openai-responses-adapter.ts:786 — `withoutNulls` now traverses arrays and strips synthetic nulls at object-property positions inside array elements; null array elements themselves are preserved because the strict rewrite never made them representable (they are model-sent data, not synthetic omission markers).

## Validation
- `pnpm vitest run` (packages/core-ai-gateway) — 220/220 passed, including a new regression test asserting `{"links":[{"label":null}],"tags":["a",null]}` → `{"links":[{}]... }` strips the object-member null and keeps the null array element.
- `tsc --noEmit` — passed
- `pnpm build` — passed

## Notes
- None declined or deferred in this pass.

sbluemin:
Thanks for the second pass. Disposition on the remaining P1:

**Declined — "Preserve real null-valued members inside arrays"**

The observation is technically accurate, but it describes the mechanism's shipped, documented trade-off rather than a regression introduced here. `withoutNulls` has always been schema-blind: a required member declared `{ "type": ["string", "null"] }` at any record position is stripped today, by design — this package's contract (AGENTS.md, "strict rewrite and null strip are one mechanism") treats property-position nulls as omission markers, because Anthropic tool semantics express "not provided" as absence, not null. This PR extends that same policy uniformly to objects inside arrays; the alternative (schema-aware stripping) requires threading per-tool schema context into response parsing, which is an architectural change out of scope for this fix.

Weighing the two failure modes: synthetic nulls inside arrays occur on real, observed paths (any tool with an array of objects carrying optional members — e.g. MultiEdit's `edits[].replace_all` — fails on the forwarded null), while a client-declared nullable required member inside an array item has not been observed through this gateway. Uniform behavior with record positions is the correct trade-off here.

**Deferred** — schema-aware null stripping is a real design idea worth recording, but it belongs to a dedicated change with its own contract update, not this 400-repair PR.

PR #566, opened 2026-08-07
feat(fleet-console)!: own remote hosts, reach them by fleet:// link

sbluemin:
## Summary

- **Console owns which other consoles this one can reach.** Settings → Remote access keeps the list: paste an access link, the server confirms the certificate over a plain socket before saving anything, and from then on each host is named, checked, opened, and forgotten from one card. The command-band chip switches between them. That list is loopback-only and stores where to go and what certificate to trust, never a credential.
- **Access links are one `fleet://join?code=…` string.** Address, credential, fingerprint, and console name travel inside one versioned envelope, so pasting a link into a chat no longer exposes a private address. Fleet Desktop registers the scheme and forwards the raw argument to the local console without reading it.
- **Fleet Desktop went back to launching the managed console.** The startup picker, the renderer shell, and the Connect to Runtime menu and tray entries are gone — choosing a console lives in the console itself. What stays is plumbing with no surface: the certificate pin, the origin admission, and the grant exchange. SSH-based remote runtimes are removed.
- The switcher also lists consoles already running on this machine, so reaching one takes no link at all. On Windows that includes a console inside a WSL distribution.

Two measurements shaped the design. Chromium caches a failed certificate verdict for the whole app run and no API invalidates it, so a host is judged over a plain `tls.connect` **before** the browser engine ever sees it. And `loadURL` resolves on a 401 body, so the console is fetched over the pinned session and required to answer as a console before the window moves.

## Test Plan

- [x] `pnpm --filter @dotobokuri/fleet-console test` — 139 files, 1,587 passed / 18 skipped. 4 failures in `launch.test.ts` (3) and `triage-store.test.ts` (1) are **pre-existing**: the identical set fails on a detached `canary` worktree at `49c1a814`.
- [x] `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` — clean
- [x] `pnpm --filter @dotobokuri/fleet-console build` — clean
- [x] `pnpm --filter @dotobokuri/fleet-desktop test` — 34 files, 230 passed
- [x] `pnpm --filter @dotobokuri/fleet-desktop exec tsc --noEmit` / `build` — clean
- [x] `pnpm check:core-boundary` — clean
- [x] `node scripts/compile-changelog-fragments.mjs --check` — 53 entries; fragment `.changelog.d/remote-access-auth-spine.md` (fleet-console + fleet-desktop)
- [x] Live, two isolated consoles: link carries no address; a forged fingerprint is refused 409, a non-`fleet://` link 400, an unreachable address 502, and the list stays unchanged in all three; the stored file holds no token; the handoff yields the grant once and `null` after; the remote listener answers 401 for `/api/v1/remote-hosts`; the shell home returns to its publisher's session only, `null` to a loopback viewer, 401 to a session-less remote one.

**Not verified:** the Windows and WSL paths — this machine runs darwin, so that code is covered by unit tests over injected seams only. Browser-side remote access stays deferred by decision; a browser cannot check the fingerprint, and the listener answers with a notice explaining what an access link looks like.


sbluemin:

All four findings were reproduced before being touched, and both P1 fixes carry a regression test.

## Addressed

- **P1 — remote session gate behind the Codex early return** (`core/host/server.ts:771`). Reproduced against a live TLS listener: `/api/v1/theaters` → 401 while `/console/codex/api/search?q=a` → **200** with a forged `Host: 127.0.0.1:<port>`. The remote admission check now runs before any routing branch; after the fix the same three paths answer 401 and loopback still answers 200.
  - The Host gate deliberately keeps its position after the Codex branch. `buildAllowedAccessSets` enumerates interfaces under a wildcard bind, so hoisting `isRequestHostAllowed` would narrow what Codex accepts in that configuration — a regression the finding did not ask for.
  - Test: `remote-access-listener.test.ts` "refuses a session-less Codex request that forges the loopback Host". Confirmed red against the previous ordering.

- **P1 — remote listener failure aborted console startup** (`core/host/server.ts:1974`). `restartRemoteAccess` already treated a bind failure as nonfatal; startup did not. Both paths now share `startRemoteAccessGuarded`.
  - Test: "still starts the console when the saved remote address no longer exists", seeding `bindHost: 203.0.113.9` (TEST-NET-3).

- **P2 — hostname bind never matched its listener** (`core/host/server.ts:1940`). `REMOTE_BIND_HOST` accepts DNS names, and `resolveListenerIdentity` compares `req.socket.localAddress`. `bindAddress` is now read from the socket after listen. Verified with a real hostname bind (`sbluemins-MacBook-Pro.local` → `172.30.1.35`): `/join` → 200, unauthenticated `/api/v1/theaters` → 401, i.e. the listener resolves and the session gate is what refuses.

- **P2 — one cookie name across ports** (`core/host/auth.ts:277`). `sessionCookieName(port)` writes the port into the name. Verified live: `Set-Cookie: fleet_console_session_59814=…`. Sessions are in-memory, so the rename costs nothing on upgrade.

## Validation

- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` / `build` — clean
- `pnpm --filter @dotobokuri/fleet-console test` — 1,590 passed. The same 4 failures in `launch.test.ts` (3) and `triage-store.test.ts` (1) reproduce on a detached `canary` worktree at `49c1a814`, so they are pre-existing and untouched by this PR.
- `pnpm --filter @dotobokuri/fleet-desktop exec tsc --noEmit` / `test` (230 passed) / `build` — clean
- `pnpm check:core-boundary`, `node scripts/compile-changelog-fragments.mjs --check` — clean

## Notes

- Nothing declined or deferred; all four were code-local defects on supported paths.
- No changelog entry was added for this pass: every fix repairs behavior introduced by this same unreleased PR, which the fragment already describes, so a standalone `Fixed` line would describe behavior no shipped version exposed.

sbluemin:

## Addressed

- **P2 — wildcard remote bind** (`core/host/settings/settings-domain.ts:52`). Confirmed: the remote listener takes the same port as loopback on a different interface, so `0.0.0.0:<port>` always loses that port to `127.0.0.1:<port>` and returns `EADDRINUSE`. The value was storable but could never produce a listener, and the surfaced error read as "another program is using that address". `0.0.0.0` now joins the loopback addresses in `isValidRemoteBindHost`, and the client validator matches so the field refuses it at entry.
  - Test: `access-registry.test.ts` "refuses %s, which can never produce a listener".

- **P2 — remote navigation outside Fleet Desktop** (`command-band-system-cluster.tsx:168`, `global-settings.tsx:749`). Confirmed on both surfaces. Opening is now offered only when the Desktop shell holds the window; the switcher row and the Settings row say `needs Fleet Desktop` / `여는 데는 Fleet Desktop이 필요합니다` instead of navigating into a 401 or a certificate error and ejecting the user from a working local console.
  - Adding, renaming, forgetting, and probing a host stay available in a browser: that work is server-side, and only the certificate handshake needs the shell. This keeps the PR's stated exclusion — browser-side remote access is deferred because a browser cannot check the fingerprint — rather than widening it.
  - The shell marker got one definition (`desktop-shell.ts:isDesktopShell`) that the two new call sites read. The three inline copies that predate this branch were left alone; converting them is a refactor this pass did not need.

## Validation

- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` / `build` — clean
- `pnpm --filter @dotobokuri/fleet-console test` — 1,595 passed. The same 4 failures in `launch.test.ts` (3) and `triage-store.test.ts` (1) reproduce on a detached `canary` worktree at `49c1a814`; pre-existing and untouched.
- `pnpm check:core-boundary`, `node scripts/compile-changelog-fragments.mjs --check` — clean

## Notes

- Nothing declined or deferred this pass.
- No changelog entry: both fixes repair behavior introduced by this same unreleased PR, which the existing fragment already describes.

sbluemin:
@codex Both findings from this pass are fixed. Closing the review loop here at the third pass.

## Addressed

- **P2 — discovered local consoles did nothing** (`remote-bridge.ts:202`). Confirmed by reading the path: `consoleTarget` recognised only the shell's own console, so every other row from `/api/v1/local-consoles` fell through to `applyWindowPolicy`, which denies anything outside the active origin — a silent dead click on a feature this PR adds. The bridge now claims loopback console targets and asks the local console whether it actually reported that origin before moving the window.
  - That question is the substance of the fix, not ceremony. A page served by a remote console could otherwise aim the window at any `http://127.0.0.1:<port>`, which resolves to a different service on the viewer's own machine — the same hazard that made `/api/v1/local-consoles` loopback-only. Opening a confirmed one needs no pin and no grant.
  - Tests: "opens another console on this machine only after the local console confirms it found one" and "refuses a loopback port the local console never reported".

- **P2 — uppercase DNS bind name** (`server.ts:1967`). Reproduced: with `bindHost` = `sbluemins-MacBook-Pro.local`, a lowercased `Host` header answered **403** while the listener and certificate were healthy. Folded once at the settings boundary, which also corrects a value already stored; the same request now answers **200** and the stored value reads `sbluemins-macbook-pro.local`.
  - Worth recording why the earlier hostname verification missed it: curl preserves the case it is given, so only a client that behaves like a browser exposes the mismatch.

## Validation

- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` / `build` — clean; `test` — 1,595 passed
- `pnpm --filter @dotobokuri/fleet-desktop exec tsc --noEmit` / `build` — clean; `test` — 233 passed
- `pnpm check:core-boundary`, `node scripts/compile-changelog-fragments.mjs --check` — clean
- The 4 failures in `launch.test.ts` (3) and `triage-store.test.ts` (1) reproduce on a detached `canary` worktree at `49c1a814`; pre-existing and untouched by this PR.

## Notes

- Three passes, eight findings, all eight fixed; nothing declined or deferred. One reviewer suggestion was implemented differently on purpose and recorded: the Host gate keeps its position after the Codex branch, because `buildAllowedAccessSets` enumerates interfaces under a wildcard bind and hoisting it would narrow what Codex accepts.
- No changelog entries for any review pass: every fix repairs behavior introduced by this same unreleased PR, which the existing fragment already describes.

PR #567, opened 2026-08-07
feat(console): start an Operation from a global prompt composer

sbluemin:
## Summary

- `Mod+J`로 어디서든 Quick Launch 컴포저를 연다. 프롬프트를 적고 Theater와 모델·추론 강도를 고른 뒤 Enter를 누르면, 이미 그 프롬프트를 첫 턴으로 제출한 Claude (Gateway) Operation이 뜬다.
- 프롬프트는 PTY 주입이 아니라 **fleet-admiral이 조립하는 `claude` argv의 후미 위치 인자**로 전달된다. 스폰 직후 PTY 주입은 텍스트만 도달하고 제출이 유실되기 때문이다(실측: create 응답 147ms → +8ms 주입 시 입력창에 남은 채 미제출 고착). 세션 liveness는 CLI readiness가 아니며, argv로 넘기면 그 경합 자체가 사라진다.
- 위치 인자는 최종 argv의 **맨 끝**이어야 한다 — `--mcp-config <configs...>`가 가변 인자라 그 뒤에 오면 삼켜진다. `agent-cli-native.test.ts`가 `args.at(-1)`/`args.at(-2)`로 이 순서를 못 박는다.
- 프롬프트는 spawn 전용이다. Operation payload에도, 브라우저 DTO에도 실리지 않는다(`prompt`는 이미 `FORBIDDEN_BROWSER_PAYLOAD_KEYS`에 있다).
- 마지막으로 고른 Theater·모델·강도를 기억한다(`fleet-console.quickLaunch.selection`). 카탈로그에서 사라진 조합은 ★기본 행으로 되돌려 서버 409를 예방한다.

세 계층으로 나뉜다: **fleet-admiral**(`prompt` 프로파일 옵션 → `promptArgs` → argv 말미 합성) · **터미널 플러그인**(`POST /sessions` body `prompt` → launch context → `resolveProfile`) · **콘솔 클라이언트**(컴포저 · Theater 칩 · 캔버스 우클릭 문법 그대로의 모델×강도 플라이아웃).

실행은 컴포저가 직접 하지 않는다. `pendingQuickLaunch`만 남기고 Operations 화면이 `activeTheaterId`가 대상과 일치할 때 소비한다(`pendingOperationFocus`와 같은 request/consume 계약) — 덕분에 대상이 비활성 Theater여도 전환·지오메트리·포커스 승계가 기존 규율을 그대로 탄다.

## Test Plan

- [x] `pnpm --filter @dotobokuri/fleet-admiral test` — 13 files / 130 tests pass
- [x] `pnpm --filter @dotobokuri/fleet-admiral build` · `typecheck` — pass
- [x] terminal plugin `vitest run` (자체 러너) — 61 files / 507 tests pass
- [x] `pnpm --filter @dotobokuri/fleet-console exec vitest run` (touched 13 files) — 160 tests pass
- [x] `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` — 0 errors
- [x] `pnpm --filter @dotobokuri/fleet-console build` — pass
- [x] `pnpm run check:core-boundary` — pass
- [x] `node scripts/compile-changelog-fragments.mjs --check` — 46 entries OK
- [x] 격리 콘솔 실브라우저 실측: `Mod+J`로 컴포저 오픈(터미널 포커스 상태 포함) · 모델×강도 플라이아웃 동작 · 실제 실행 시 `Sonnet 5 with low effort`로 스폰되어 프롬프트가 첫 제출 턴으로 들어가고 응답 완료
- [x] 실측 검증: `invalid_prompt` 400 · `prompt_too_long` 400 · 생성 응답과 Operation payload 모두 `prompt` 키 없음

## Changelog

- `.changelog.d/quick-launch-composer.md` (`### fleet-console` / `#### Added`)


sbluemin:

## Addressed

- **P1 — cmd.exe expansion (`injection.ts:225`)** — FIXED, and the repository already codifies this hazard: `core-process/src/bin-resolver.ts` wraps Windows `.cmd`/`.bat` shims as `cmd.exe /d /s /c call <shim>`, and the adjacent `rejectCmdExpansionSensitiveShim` *throws* when the shim path carries `%` or `^`. Escaping is not available on a `cmd` command line (`%%` is collapsed only inside batch files), so the prompt now follows the same rule the repo already chose for that path: refuse rather than escape. `assertLaunchPromptShimSafe(prompt, prefixArgs)` (`packages/fleet-admiral/src/agent-cli/prompt.ts`) only engages when `prefixArgs` is non-empty — i.e. when this launch actually goes through `cmd.exe` — so POSIX and direct-executable launches are unconstrained. The typed `LaunchPromptUnsafeError` is mapped by the session route to `400 prompt_unsafe_for_shim` rather than the generic `503`.
- **P2 — prompt limit enforced after the composer closes (`quick-launch.tsx:107`)** — FIXED. The composer now holds the limit itself: submission is blocked over `QUICK_LAUNCH_PROMPT_MAX_CHARS`, the composer and its draft stay open, and an inline status reports how many characters over it is. Because browser code cannot import the Node package that owns the server constant, the browser copy is a duplicate — so `tests/quick-launch.test.ts` reads `MAX_LAUNCH_PROMPT_CHARS` out of the fleet-admiral source and pins the two values equal, which is what makes the duplication safe.

## Also changed (found by internal review, same pass)

- `quickLaunchOpen` added to the What's New suppress list (`whatsnew-modal.tsx`) — without it an auto-opened What's New and the composer could stack two `aria-modal` dialogs and fight over Escape/focus.
- The composer overlay's `prefers-reduced-motion` clause is now pinned in `instrument-design-contract.test.ts`, matching the `.fc-select__*` precedent.

## Validation

- `pnpm --filter @dotobokuri/fleet-admiral test` — 14 files / 135 tests pass (5 new tests cover the cmd-wrapped refusal, including the non-wrapped path staying unconstrained)
- `pnpm --filter @dotobokuri/fleet-admiral build` — pass
- terminal plugin `vitest run` — 61 files / 507 tests pass
- `pnpm --filter @dotobokuri/fleet-console exec vitest run` (13 touched files) — 162 tests pass
- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` — 0 errors
- `pnpm run check:core-boundary` — pass
- `node scripts/compile-changelog-fragments.mjs --check` — 47 entries OK

## Notes

- The Windows behaviour itself was not executed here (this host is macOS). What was verified directly is the mechanism in code — the `cmd.exe` wrapper and the pre-existing refusal doctrine it already applies to the shim path — plus unit tests that drive the guard with a cmd-wrapped `prefixArgs` shape.
- Failure feedback for the *non-deterministic* rejections (`409 gateway_model_not_enabled`, agent CLI unavailable, and the new shim refusal) is deliberately not in this PR. Surfacing those requires the composer to await the launch result, which would undo the request/consume handoff this PR chose so that launch geometry and focus inheritance stay owned by the Operations view. It is tracked as the already-planned P2 rollout item.

sbluemin:

## Addressed

- **P1 — reject all cmd metacharacters (`prompt.ts:6`)** — FIXED, and the citation was the decisive part: `quoteForCmd` in `core-unified-agent/src/connection/BaseConnection.ts` classifies `" & < > ( ) @ ^ |` as cmd-special *and its own doc comment requires doubled inner quotes plus `windowsVerbatimArguments`*. The node-pty launch path supplies neither, so the assumption that the positional is safely inside quotes never held. The guard now refuses that full set plus `%`, and the tests drive each character individually and assert the non-wrapped path stays unconstrained.
- **P2 — `FLEET_TERMINAL_CMD` discards the prompt (`launch.ts:156`)** — FIXED by making prompt-based launch unavailable in that configuration, which was your second suggested option. The override replaces the Agent CLI wholesale (model, effort, resume, system prompt, MCP config and plugin dirs are all dropped with it), so there is no argv contract to append user text to; appending it would be the worse hazard. A prompt-carrying launch now fails with `400 prompt_unsupported_launch` instead of opening an Operation that was never told anything.
- **P2 — stale catalog enables submission (`quick-launch.tsx:59`)** — FIXED at the root rather than at that trigger. The underlying harm you identified in both passes is the same one: the composer closes on submit and never learns the outcome, so *any* server rejection loses the draft. A rejected launch now travels back through the same request channel and reopens the composer with the text intact. That covers a model disabled since the last open, an unavailable Agent CLI, and both refusals above — not only the length case the client can predict on its own. The client-side length gate from the previous pass stays, because that one is knowable before sending.

## Validation

- `pnpm --filter @dotobokuri/fleet-admiral test` — 14 files / 150 tests pass
- `pnpm --filter @dotobokuri/fleet-admiral build` — pass
- terminal plugin `vitest run` — 61 files / 507 tests pass
- `pnpm --filter @dotobokuri/fleet-console exec vitest run` (13 touched files) — 164 tests pass
- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` — 0 errors
- `pnpm run check:core-boundary` — pass
- `node scripts/compile-changelog-fragments.mjs --check` — 47 entries OK

## Notes

- Windows itself was not executed here (macOS host). What is verified directly is the mechanism in code — the `cmd.exe` wrapper, the repository's own cmd-special classification, and the pre-existing refuse-don't-escape doctrine in `core-process` — plus unit tests driving the guard with a cmd-wrapped `prefixArgs` shape.
- The refusal is deliberately a refusal rather than an escape: a `cmd` command line does not collapse `%%`, so there is no escaping scheme available on this path, and `core-process` already answers the same hazard for the shim path by throwing.

sbluemin:
@codex Addressed in `3aff751f`.

## Addressed

- **P2 — surface the launch rejection when restoring the draft (`operations.tsx:365`)** — FIXED. You were right that the recovery handler swallowed the reason, which left deterministic rejections (`prompt_unsupported_launch`, the Windows shim refusal) repeating identically on every Run. The agent-session client now throws the server's `error` code instead of a bare status line — matching the shape `setAgentCliPath` and the goal API already use in that package — the code rides back with the draft, and the composer names what to change. Unknown codes fall back to a generic reason so the composer is never silent about a rejection it cannot classify. Both locales carry every mapped key, pinned by a test.

## Validation

- `pnpm --filter @dotobokuri/fleet-console exec vitest run` (13 touched files) — 168 tests pass
- terminal plugin `vitest run` — 61 files / 507 tests pass
- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit` — 0 errors
- `pnpm run check:core-boundary` — pass
- `node scripts/compile-changelog-fragments.mjs --check` — 47 entries OK

## Closing the review loop

This was the third review pass. Every finding across all three passes was accepted and fixed — nine in total, none declined — so there is no outstanding disagreement to record. Two of them reversed an earlier position of mine, which is worth stating plainly: I first declined the `FLEET_TERMINAL_CMD` case as consistent-with-the-override, and you and a second reviewer independently showed that the launch kind stays available, so the user gets an Operation that was never told anything; and I first deferred the failure surface, then adopted it once three separate findings turned out to be the same draft-loss root cause.

Two deliberate behaviour limits are now in place and are intended, not oversights: on a Windows `.cmd` shim a prompt containing `" & < > ( ) @ ^ | %` is refused rather than escaped (a `cmd` command line has no escaping scheme for `%`, and `core-process` already answers this hazard by throwing), and a prompt-carrying launch is refused under `FLEET_TERMINAL_CMD`. Both now tell the user which one applied.

Proceeding to merge after a final audit of the cumulative review diff against the frozen scope.

PR #580, opened 2026-08-08
feat(console): move the AI Gateway spend order onto each provider row

sbluemin:
## Summary
- AI Gateway 설정의 별도 '공급자 우선 소진' 칩 줄을 없애고, 각 공급자 헤드의 글리프·이름 오른쪽에 순위를 표시하는 토글 버튼을 둡니다.
- 토글에 따른 순위 조정은 기존 문법 그대로입니다: 꺼진 공급자를 켜면 순서 끝에 추가되고, 번호가 붙은 토글을 다시 누르면 빠지며 나머지 순위가 당겨집니다. 저장 경로(`providerPriority` 에코 계약)는 손대지 않았습니다.
- 칩 줄이 지고 있던 라벨·도움말 대신, 토글에 hover 하거나 키보드 포커스를 주면 소진 순서의 의미를 한 줄로 요약하는 말풍선이 열립니다.

## Test Plan
- [x] `pnpm --filter @fleet-plugins/terminal typecheck` — 통과
- [x] `pnpm --filter @fleet-plugins/terminal test` — 62 files / 517 tests 통과 (칩 테스트를 토글 테스트로 재작성)
- [x] `pnpm --filter @dotobokuri/fleet-console typecheck` — 통과
- [x] `pnpm --filter @dotobokuri/fleet-console vitest run tests/instrument-design-contract.test.ts` — 41 tests 통과 (토글 문법·말풍선 키보드 도달성 계약 추가)
- [x] 헤디드 실측(격리 콘솔): 헤드 배치(글리프 450 → 이름 488 → 토글 546 → 부제 우측), Cursor→Codex 토글 시 순위 1/2 및 `ai-gateway.json` = `["cursor","codex"]`, 재클릭 제거 후 `["codex"]`, 리로드 후 유지, hover·Shift+Tab 포커스 양쪽에서 말풍선 표시, instrument/light/maritime/carbon 4테마 스크린샷
- [x] 변경 로그: `.changelog.d/gateway-priority-head-toggle.md`


sbluemin:

## Addressed
- P2 a11y (index.tsx:811) — FIX. The tip is no longer `aria-hidden`; the toggle now points at it with `aria-describedby` (one id per provider), so screen readers get the meaning of the spend order alongside the action-only accessible name.
- P2 narrow header (agent-cli.css:356) — FIX, reproduced first. At 375px the head measured `scrollWidth 359` vs `clientWidth 253` (106px overflow, the subtitle clipped; the page itself did not scroll horizontally). `.ai-gateway-provider-head` now wraps with a row gap, so the subtitle folds to a second line: overflow measured 0 after the fix, and the 1280px layout is unchanged (head height 26, glyph 450 / name 488 / toggle 546 / subtitle 1053).
- P2 tip clipping (agent-cli.css:398) — FIX, reproduced first. At 375px the tip measured right edge 415 against a 375 viewport (40px past it). The tip is now anchored to the provider head instead of the toggle and capped at `max-width: 100%`, so its width can never start mid-row: measured right edge 309 at 375px, and 334px wide on one line at 1280px.

## Validation
- pnpm --filter @fleet-plugins/terminal typecheck — pass
- pnpm --filter @fleet-plugins/terminal test — 62 files / 517 tests pass
- pnpm --filter @dotobokuri/fleet-console typecheck — pass
- pnpm --filter @dotobokuri/fleet-console vitest run tests/instrument-design-contract.test.ts — 41 tests pass (contract extended: sibling hover/focus selector, head as the tip's containing block, head wrapping)
- pnpm --filter @dotobokuri/fleet-console build — pass
- Headed measurement on an isolated Console at 375x812 and 1280x800, hover and keyboard focus both open the tip.

## Notes
- The tip now aligns to the head's left edge rather than the toggle's. That is the deliberate cost of making viewport overflow structurally impossible without anchor positioning; the tip still sits directly under its own provider row.

sbluemin:
@codex On the changelog finding — deferred, with the reason recorded rather than silently skipped.

The norm you cite is right: this redesign touches behavior no public release exposed, and `.changelog.d/AGENTS.md:45` says to fold it into the pending entry. That fold cannot land in this PR, though. `.github/workflows/changelog-fragment.yml` compares every fragment present on the merge base and fails the PR if any of them differ by M, D or R:

    base_fragments=$(git ls-tree -r --name-only "$merge_base" -- .changelog.d | fragment_filter)
    ... if ! git diff --quiet "$merge_base" "$HEAD_REF" -- "$fragment"; then disturbed=...

`.changelog.d/feat-gateway-priority-ui.md` is on the merge base (verified with `git ls-tree`), so editing it here trips that gate. The other exit is the `no-changelog` label, which additionally requires `No-Changelog-Reason:` to be one of `internal-refactor | boundary-gate | doctrine-prompt | test-repair | release-tooling` — none of which is true for a user-visible settings control, and I will not file a false reason code to close a comment.

So the remaining path is a direct `canary` commit that merges the two fragments (the guard is `pull_request`-only), and that is the repository owner's call, not this PR's. The standalone fragment is worded as the increment this PR adds rather than a restatement of the pending bullet, so the two compile as consistent lines, not contradictory ones — the cost is redundancy at release time, which the owner has been told about explicitly.

No other findings in this pass, so this PR proceeds to the final context audit and merge.

PR #593, opened 2026-08-09
fix(fleet-console): stop a resident agent from wedging an Operation in background

sbluemin:
## Summary
- A named agent stays registered with the Claude session after it finishes so it can be messaged again, and every later stop hook still lists it as `running`. The payload carries no field separating resident from busy, so each turn end re-read that entry as live work and the Operation stayed in `BACKGROUND` — never reaching idle, and so never promoted to `AWAITING` — with nothing actually running.
- The session now remembers the agent ids whose own stop hook already arrived and excludes them from later reports, instead of discarding that knowledge after a single request. Remembered ids are pruned by the next fully readable task list that no longer carries them, so the memory stays bounded by the live list; it survives a cleared badge and is dropped with the session lifecycle.
- Workflows, freshly dispatched agents, unreadable reports, and shell tasks keep their existing behavior on this axis.

Changelog-Amend: background-pending-live-tasks.md

The background-pending axis this corrects is itself unreleased, so the note is folded into its pending fragment rather than emitted as a standalone `Fixed` entry.

## Test Plan
- [x] `pnpm --filter @fleet-plugins/terminal typecheck`
- [x] `pnpm --filter @fleet-plugins/terminal build`
- [x] `pnpm --filter @fleet-plugins/terminal test` (558 passed)
- [x] Workspace `pnpm typecheck` and `pnpm test` (the three failures seen locally are host-env artifacts in `core-infra` / `fleet-desktop` and pass with the `FLEET_*` data-dir variables unset)
- [x] Negative control: the new route-level regression test fails when the remembered ids are not passed through


sbluemin:
@codex Thanks — the observation is correct, but the prescribed fix is declined because the signal it requires does not exist.

## Declined — P2 "Re-arm settled residents when they receive new work" (`background-report.ts:98`)

**The factual claim is confirmed.** A resident that is re-tasked after its first `SubagentStop` stays excluded, so its next run is under-reported on the background axis. That is a real, deliberate trade-off, and it was recorded in the PR body before this review.

**The prescription is not implementable.** "Clear or version the settled marker when a resident starts another assignment" needs an event that identifies the resident at re-dispatch. Verified against the Claude Code 2.1.226 binary:

- A re-task is delivered by appending to the task's `pendingUserMessages`, and the resident's runner picks it up from a **polling loop inside the same `runAgent` invocation** (`[inProcessRunner] ... found pending user message (poll #N)`). `SubagentStart` is emitted only in the agent-creation path, which that loop never re-enters — so no start-side hook fires on re-dispatch.
- `TeammateIdle`, `TaskCreated`, and `TaskCompleted` carry `teammate_name` / `team_name` / `task_id` — never the `agent_id` that `background_tasks[].id` is keyed by — so none of them can re-arm a marker.
- The payload cannot substitute for the event. `status` is `"running"` for a resident whether it is busy or idle (the CLI's own busy test reads an `isIdle` field that is not serialized into the hook payload), and teammate entries carry no extra fields at all.

**Versioning on a heuristic would be worse than the gap.** The only remaining lever is guessing from `description`, which is an unverified runtime assumption about a field whose update semantics are not established.

**Clearing the marker at turn start was considered and rejected:** it fixes the re-task case but reintroduces the exact defect this PR closes — the resident is still listed at the turn end that follows its own stop notification, so the Operation would wedge in `BACKGROUND` again.

**Direction of failure was chosen deliberately.** Under-reporting costs a missing badge while a re-tasked resident runs. Over-reporting — the shipped-in-this-branch behavior — wedges the whole idle/awaiting/dormant axis for up to the 30-minute TTL with nothing running, and blocks idle dormancy the entire time. That is the bug this PR was opened for. Closing the remaining gap needs an upstream signal (a start-side hook carrying `agent_id`, or `isIdle` in the payload), not a local heuristic.

PR #594, opened 2026-08-09
feat(fleet-console): separate Status recovery shelves

sbluemin:
## Summary
- split each Theater's Alt+S Status view into four live activity sections plus dedicated Minimized and Dormant recovery shelves
- preserve distinct restore and resume actions while keeping inactive Theater status visible without mounting foreign Operation bodies
- retain arrival signals and first-entry boot minimization behavior across recovery flows

## Test Plan
- [x] `vitest` affected suite: 7 files, 125 tests passed
- [x] Fleet Console `typecheck`
- [x] Fleet Console production `build`
- [x] `git diff --check`
- [x] isolated browser E2E for active/inactive Theater shelves, restore, resume, body-mount boundary, and session-only Status axis
- [x] changelog fragment: `.changelog.d/product-war-room-status-proposal.md`


sbluemin:

## Addressed
- stale boot-minimization protection after an interrupted foreign-Theater resume — `operations.tsx` now distinguishes source, target, and third-Theater transitions and consumes the protection only for the destination effect
- interrupted A -> B -> C transition — integration coverage now verifies a later first B entry receives normal boot minimization

## Validation
- affected Vitest suite: 7 files, 126 tests passed
- Fleet Console typecheck: passed
- Fleet Console production build: passed
- changelog compiler: 113 entries validated
- `git diff --check`: passed

## Notes
- preserved the frozen product contract: dormant resume does not set focus state, foreign bodies mount only after the owning Theater switch, and same-Theater resume does not reload the canvas

sbluemin:
@codex Please re-review the final context-audited head.

## Review disposition update
- DECLINED: stale resume-protection mismatch hardening. The reported A -> B -> C mismatch required two synchronous internal store updates in one `act()`; normal user events flush the prior passive effect, and no supported app path was found that performs the second Theater switch before that effect.
- ROLLED BACK: commit `a3c3a682` via `796b74ef` rather than stacking hypothetical hardening onto the verified implementation.

## Preserved product contract
- foreign dormant resume switches to the owning Theater before its body mounts and does not set focus state
- the current destination resume target is excluded from first-entry minimization
- same-Theater resume does not reload the canvas
- no cross-Theater body mounting was added

## Validation
- affected Vitest suite: 7 files, 125 tests passed
- Fleet Console typecheck: passed
- Fleet Console production build: passed
- cumulative review-driven content diff from `REVIEW_BASE_HEAD` is empty

PR #638, opened 2026-08-11
feat(fleet-console): reach this console through a NAT endpoint

sbluemin:
## Summary

원격 엔드포인트를 **수신 튜플**(`listenAddress`/`listenPort`)과 **공표 튜플**(`advertisedHost`/`advertisedPort`)로 분리해, 라우터 뒤의 콘솔을 공개 호스트 이름으로 알릴 수 있게 합니다. 인증서 신원은 공표 호스트에 묶이고, 리스너는 실제 수신 튜플에 바인드합니다.

설정 카드는 **초안/적용 트랜잭션**입니다. 인터페이스 선택·포트 모드 변경·공개 엔드포인트 토글 어느 것도 실행 중인 리스너를 건드리지 않으며, 저장은 `수신 시작`·`나중을 위해 저장`·`변경 사항 적용`·`수신 중지` 넷뿐입니다. 적용 전에 무엇을 잃는지 먼저 말합니다 — 재시작은 페어링을 유지하지만, 기기가 신뢰하던 주소가 바뀌면 세션·미사용 링크·전 기기 페어링이 사라집니다.

원격 리스너의 무인증 표면을 **`POST /api/v1/join` 하나로** 줄이고, 그 문에 출처별 실패 예산과 동시 상한을 두었습니다. 리스너 자신의 설정은 소유자 쪽에 남습니다 — 원격 세션에는 Remote access 섹션이 아예 실리지 않습니다.

## Test Plan

- [x] `pnpm exec tsc --noEmit -p tsconfig.json` — 0 오류
- [x] `pnpm exec vitest run` — 1909 passed / 24 skipped, 실패 0
- [x] `node scripts/compile-changelog-fragments.mjs --check` — OK: 8 entries
- [x] 실 NAT 환경 e2e — 공개 호스트 이름 + 포트 포워딩으로 실제 기기 페어링 성공, 인증서 지문 일치
- [x] 업그레이드 재현 — canary 형식(`bindHost` + `listener.json` v1)에서 부팅 시 origin·지문·기기 목록 보존, 구버전 페어링 쿠키만으로 재개 204
- [x] 예산 발동 — 실패 11회째 429 + `Retry-After`, 창 만료 후 정상 페어링 204
- [x] 브라우저 표면 제거 — 원격 리스너에서 `/`·`/join`·`/console/`·`/api/v1/state` 모두 401
- [x] 섹션 은닉 — 원격 시야에서 Remote access 내비 항목·DOM 부재, `?section=remote-access` 직접 진입도 차단 / 루프백에서는 정상 렌더


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed

- **P1 — Keep the saved Auto port when it is occupied** — `core/host/server.ts:2411,2431` — the Auto retry loop now only searches for a replacement before an address has been published. Once `remote/listener.json` remembers an endpoint, an occupied port closes the socket it just opened and throws `FLEET_REMOTE_PORT_UNAVAILABLE` instead of sliding. This is the path that error code and its user-facing message were written for; nothing reached it before.
- **P2 — Close a newly bound listener if reconciliation storage fails** — `core/host/server.ts:2386` — `remoteServer` now takes ownership of the started server before the revocation writes, so a throw from those unguarded `fs` calls reaches `stopRemoteAccess` with something to close.
- **P2 — Return settings after listener reconciliation** — `core/host/settings/settings-domain.ts:516` — the mutation result is rebuilt from the store after `onRemoteAccessChanged`, so an Auto-fallback rewrite inside that callback is what the client receives.
- **P2 — Generate a valid candidate when switching to Auto** — `core/client/src/pages/global-settings.tsx:869` — the mode transition rerolls when the retained value falls outside 49152–65535, so a Custom `443` no longer produces a ready-looking card that the server rejects.

## Validation

- `pnpm exec tsc --noEmit -p tsconfig.json` — 0 errors
- `pnpm exec vitest run` — 1911 passed / 24 skipped, 0 failed
- New regression test occupies a published LAN-only Auto port across a restart and asserts the listener fails with `remote_port_unavailable` while the pairing survives. Verified it fails without the fix (not a vacuous pass).

## Notes

- Nothing declined or deferred in this pass. P1 was accepted on the strength of the repository's own invariant (`runtime/fleet-console/AGENTS.md:31`, "it never slides to another port on its own") rather than severity alone.
- First-start behavior is unchanged: with no published endpoint, Auto still searches for a free candidate and LAN-only still persists it. The existing "keeps a successful LAN-only Auto retry active" test covers that path and still passes.

sbluemin:

## Addressed

- **P1 — Revoke old credentials before a listener bind can fail** — `core/host/server.ts:2375` — the identity decision and the revocation now run before the bind. This is only sound because of the previous pass: a published endpoint no longer slides to another port, so the port that will open is already known at that point, and before an address is published the port axis is not part of the decision. Regression test renames the advertised host, occupies the published port, and asserts the device list is empty after the failed start; it fails without the change.
- **P2 — Close the Auto fallback listener when persistence fails** — `core/host/server.ts:2447` — the fresh LAN-only fallback closes the socket it just opened if the settings write throws, before rethrowing.
- **P2 — Reject unusable advertised hosts in the client** — `core/client/src/types.ts:271` — the client validator now mirrors the listen-address set. This divergence was introduced when the server side was tightened in an earlier commit without updating the client mirror; the unit test that asserted the old, looser behavior was updated with it.

## Accepted in part

- **P2 — Reset the join guard for each new listener** — the observation is right and the prescription is not. The counter said "since the listener started" while counting from process start, so **the wording is fixed** (`settings.remote.rejectedJoins`, plus the type and call-site comments). The **lifetime is deliberate and unchanged**: a budget cleared by a listener restart is a budget with a reset button, and no supported flow needs that reset. A legitimate device is already protected because a successful pairing clears its source's budget, and the owner cannot be locked out — the loopback listener is unaffected.

## Validation

- `pnpm exec tsc --noEmit -p tsconfig.json` — 0 errors
- `pnpm exec vitest run` — 1912 passed / 24 skipped, 0 failed
- Both new regression tests verified to fail without their fix (not vacuous passes).

sbluemin:
@codex Third pass addressed. Closing the review loop here under the workflow's stop rule (three passes), with the final context audit as the merge gate.

## Addressed

- **P1 — Canonicalize the default HTTPS port in listener identities** — `core/host/auth.ts:319`, `core/host/server.ts:809` — reproduced first: with `advertisedPort: 443`, a request carrying `Host: nat.example.test` (what Node, Electron and browsers all send) answered `403 host_mismatch`, while `Host: nat.example.test:443` reached `401`. Authority and origin are now built with standard default-port normalization, and the incoming Host is normalized the same way so both spellings of one authority resolve together. The numeric port is deliberately untouched — pairing cookies are named after it, and normalizing there would rename the door devices come back through. A regression test asserts both spellings pass the gate and that the cookie is still `fleet_console_session_443`; it fails without the change.
- **P2 — Clear the published-port flag after identity revocation** — `core/host/server.ts:2396` — the flag now reflects the state after the revocation. This was a consequence of the first pass's fix, so it is narrowed to the case that fix was for — a live published address — rather than widened.

## Validation

- `pnpm exec tsc --noEmit -p tsconfig.json` — 0 errors
- `pnpm exec vitest run` — 1913 passed / 24 skipped, 0 failed
- `node scripts/compile-changelog-fragments.mjs --check` — OK
- Every regression test added across the three passes was verified to fail without its fix.

## Notes

- Across three passes: 9 findings fixed, 1 accepted in part (the join guard's wording was wrong, its process-scoped lifetime is deliberate — a budget cleared by a listener restart is a budget with a reset button). Nothing else declined.
- Two of the findings were consequences of an earlier pass's fix; both were resolved by narrowing that fix to its intended case rather than stacking a compensating change on top.

PR #650, opened 2026-08-12
feat(fleet-mobile): sign releases and distribute them to testers

sbluemin:
## Summary
- Replace the blanket refusal of Android release assembly with opt-in signing driven by the keystore environment: a release either carries a real key or does not build. The refusal lived in four places — the build script, the app-level Gradle plugin, the native module's config plugin, and a contract test pinning it — and release policy now belongs to the app-level plugin alone.
- Carry the debug security contract over to release builds and add release-only gates: a debuggable APK and the shared Android debug key both fail verification. The promoted manifest records the signing certificate digest, so a key change is visible before it ships.
- Distribute release-signed APKs to Firebase App Distribution, and give the mobile shell the independent versioning the desktop shell already has (`mobile-v*` marker tag, change detection scoped to `runtime/fleet-mobile`, feat/patch bump rules) so `main` pushes distribute only when the mobile path actually changed.

`app.json` — not `package.json` — is the version the APK carries, and nothing was bumping it. Without that, every distribution would reuse `versionCode` 1 and silently replace the previous App Distribution release instead of notifying testers. The release commit now stages `app.json` too.

Unlike the desktop bootstrap, a missing `mobile-v*` marker falls back to the previous release tag rather than forcing a build, because the first release-signed APK already went out by hand.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-mobile test` — 22/22
- [x] `pnpm --filter @dotobokuri/fleet-mobile typecheck` — 0 errors
- [x] Release build with a throwaway keystore, then with the real one: 96.8MB, not debuggable, one non-debug signer, permission and exported-component contract identical to debug
- [x] Release build refused with no keystore environment (script gate), and the Gradle guard is separately in place
- [x] Debug build rebuilt and re-verified after the shared-code refactor — no regression
- [x] Distributed to Firebase App Distribution and confirmed the release in the console
- [x] Version resolution simulated against the real repo state: mobile path changed -> 0.2.0 / versionCode 2; untouched mobile path -> no build
- [ ] `actionlint` not run — the npm wrapper ships no binary; workflow YAML was parsed but not statically linted

### DIFF
--- a/.github/workflows/mobile-release.yml
+++ b/.github/workflows/mobile-release.yml
@@ -1,0 +1,88 @@
+name: Mobile Release
+
+# stable-release.yml이 릴리스 커밋을 만든 뒤 같은 런에서 호출한다(workflow_call).
+# 산출물은 GitHub Release에 붙지 않는다. 릴리스 서명 APK는 Firebase App Distribution으로만 나가며,
+# 받는 사람은 테스터 그룹에 등록된 계정뿐이다.
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description: 빌드할 릴리스 커밋의 git ref/sha (app.json 버전이 이미 bump된 상태)
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 빌드할 git ref (미지정 시 기본 브랜치)
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  distribute:
+    name: Build and distribute Android
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+      - uses: pnpm/action-setup@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: pnpm
+      # resolveJavaHome()은 CI에서 JAVA_HOME을 쓴다(macOS 로컬에서만 Android Studio 번들 JDK를 찾는다).
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '17'
+      - uses: android-actions/setup-android@v3
+      - name: Install the pinned Android platform and build-tools
+        run: sdkmanager --install "platforms;android-36" "build-tools;36.0.0"
+      - run: pnpm install --frozen-lockfile
+
+      # 서명 자료는 파일로만 존재하고 저장소에는 들어가지 않는다. 러너는 일회용이므로 작업 디렉터리
+      # 밖(RUNNER_TEMP)에 쓰고 경로만 빌드에 넘긴다.
+      - name: Materialize the release keystore
+        env:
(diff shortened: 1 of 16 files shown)


sbluemin:

## Addressed
- P1 — `.github/workflows/stable-release.yml:228` — Confirmed on real history: the `v1.55.0` release commit edits `runtime/fleet-mobile/package.json` **and** `runtime/fleet-mobile/modules/fleet-console-view/package.json`, so the next release would have seen a mobile change and distributed an APK. Detection now excludes release commits (`git log --invert-grep --grep='^chore(release):'`) instead of excluding the files. Excluding `package.json` by pathspec was rejected because it would also hide dependency edits that genuinely change the APK.
- P2 — `runtime/fleet-mobile/scripts/lib/android-promote.mjs:132` — The promoted release manifest now carries `signerSha256`, and verification fails when it disagrees with the APK. Debug manifests still carry no signer field (all debug builds share the one Android debug key). `DISTRIBUTION.md` was corrected to describe manifest comparison rather than log output.

## Validation
- Reproduced P1 on `44762f38^..44762f38`: old pathspec reports a change, release-commit exclusion reports none; a real mobile range (`v1.54.1..v1.55.0`) still resolves to changed with a `feat` bump.
- Release rebuild — manifest carries `signerSha256: f8e5bb7d…`, matching the apksigner digest.
- Negative test — tampering `signerSha256` in the manifest makes `verify-android-release` fail; restoring it passes.
- Debug verification re-run against the new comparison — passes, no signer field.
- `pnpm --filter @dotobokuri/fleet-mobile test` — 23/23; `typecheck` — 0 errors; workflow YAML parses.

## Notes
- `actionlint` still not run: the npm wrapper ships no binary.

sbluemin:

## Addressed
- P1 — `.github/workflows/stable-release.yml:227` — Correct: with no marker, using the previous release tag as the baseline made a failed first distribution unrecoverable, because that failed release's own tag became the next baseline. Without a marker the job now always distributes (the desktop rule), and `LATEST_TAG` scopes only the feat/patch decision. The bootstrap still bumps rather than shipping the seed version, since 0.1.0 / versionCode 1 already reached testers by hand and re-uploading it would replace that release.
- P2 — `runtime/fleet-mobile/DISTRIBUTION.md:49` — Correct, and it was my own inconsistency: the export still named `$HOME/.fleet` after the keystore guidance moved to `~/.keystores`. Following the document verbatim would have failed `requireReleaseKeystore()`. Fixed to the path step 1 creates.

## Validation
- Simulated all four cases against the real repo state: no marker -> distribute (minor); marker + this PR's changes -> distribute (minor); marker + untouched mobile path -> no distribution; **no marker with the release tag advanced past the failure -> still distributes**, which is the retry the previous revision lost.
- `pnpm --filter @dotobokuri/fleet-mobile test` — 23/23; `typecheck` — 0 errors; workflow YAML parses.

## Notes
- `actionlint` still not run: the npm wrapper ships no binary.

PR #653, opened 2026-08-12
feat(fleet-console): name the running model in the command band

sbluemin:
## Summary

- 커맨드 밴드의 Operation 속성 칩이 실행 경로 이름(`Claude (Gateway)`) 대신 실제로 돌고 있는 모델 이름을 답니다. 게이트웨이 행은 `models.json`의 `name`(예: `GPT-5.6-Sol-Fast`, `K3-1M`), 네이티브 Claude 행은 `Opus`/`Fable`/`Sonnet`입니다.
- 이름의 출처는 실행 메뉴·Quick Launch 칩과 같은 실행 카탈로그 행 라벨(`bareModelName()`) 하나입니다. 라벨을 Operation payload에 복제하지 않아 이미 저장된 Operation도 그대로 반영되고, 카탈로그가 바뀐 뒤 옛 payload가 스테일 문자열을 안고 남지 않습니다.
- 밴드에서 열리는 Operation 드롭다운의 meta도 같은 색인을 씁니다 — 한 표면에서 칩과 메뉴가 서로 다른 이름을 말하지 않게 합니다.
- `launchModel`이 없는 옛 Operation과 카탈로그가 모르는 좌표는 기존 CLI 라벨로 물러납니다(회귀 없음).

## Test Plan

- [x] `pnpm --filter @dotobokuri/fleet-console typecheck` — 통과
- [x] `pnpm vitest run tests/command-band-switcher.test.tsx tests/command-band-v2-guards.test.ts` — 31건 통과 (색인·폴백 케이스 신규 포함)
- [x] `node scripts/compile-changelog-fragments.mjs --check` — 7건 검증 통과
- [x] 격리 콘솔 + 헤드리스 브라우저 실측: `codex--gpt-5.6-sol-fast` → `GPT-5.6-Sol-Fast`, `kimi--k3` → `K3-1M`, `opus[1m]` → `Opus`, `launchModel` 없음 → `Claude (Gateway)`. 밴드와 Operation 메뉴 양쪽 동일, 페이지 에러 0건

### DIFF
--- a/.changelog.d/command-band-model-name.md
+++ b/.changelog.d/command-band-model-name.md
@@ -1,0 +1,8 @@
+---
+branch: command-band-model-name
+---
+
+### fleet-console
+#### Changed
+- Name the model an Operation is actually running in the command band and its Operation menu, so a breadcrumb reads GPT-5.6-Sol-Fast or Opus instead of the launcher it went through.
+  ko: 커맨드 밴드와 Operation 메뉴가 실행 경로 이름 대신 실제로 돌고 있는 모델 이름을 답니다. 브레드크럼에 GPT-5.6-Sol-Fast나 Opus가 그대로 보입니다.
--- a/runtime/fleet-console/core/client/src/components/command-band-guards.ts
+++ b/runtime/fleet-console/core/client/src/components/command-band-guards.ts
@@ -1,3 +1,5 @@
+import type { OperationCatalogPlugin } from "@fleet-console/sdk/operations";
+
 import { flattenGroupedOrder } from "../store.js";
 import type { OperationGroup, OperationNode } from "../types.js";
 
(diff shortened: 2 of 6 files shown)


sbluemin:

## Addressed
- P2 catalog fetch retry — `runtime/fleet-console/core/client/src/components/command-band.tsx:239` — dropped the ref-based request ledger instead of adding a reschedule. The ledger was the cause: under StrictMode the second setup saw the key left by the aborted first setup and backed off, and the late `delete` had no render to act on. The effect dependency already limits the fetch to a changed coordinate, so the ledger only bought a skipped refetch when the active Operation toggles between a known and an unknown coordinate — the same cost Quick Launch already pays on every open.

## Validation
- `pnpm --filter @dotobokuri/fleet-console typecheck` — pass (0 errors)
- `pnpm vitest run tests/command-band-switcher.test.tsx tests/command-band-v2-guards.test.ts tests/command-band-focus.test.ts` — 34 passed
- Headless browser against an isolated console, with the first `/api/v1/operations/catalog` request forced to reject: `codex--gpt-5.6-sol-fast` -> `GPT-5.6-Sol-Fast`, `launchModel`-less Operation -> `Claude (Gateway)`, toggling back -> `GPT-5.6-Sol-Fast`. No page errors. Same result before and after the fix on the production bundle, which does not double-invoke effects; the StrictMode path is the development channel, and this repo already carries the same trap at `add-host-dialog.tsx:41`.

## Notes
- No behavior change to the fallback contract: an Operation without `launchModel`, or one whose coordinate the catalog does not know, still shows the CLI label.

sbluemin:

## Addressed
- P2 refetch on resolve — `runtime/fleet-console/core/client/src/components/command-band.tsx:242` — the effect now depends on the active coordinate itself, and a guard skips the request when the index already knows that coordinate (or when a coordinate-less Operation is active and the index is already populated). A successful lookup no longer becomes its own next trigger, and switching to an already-indexed model no longer refetches either.

## Validation
- `pnpm --filter @dotobokuri/fleet-console typecheck` — pass (0 errors)
- `pnpm vitest run tests/command-band-switcher.test.tsx tests/command-band-v2-guards.test.ts tests/command-band-focus.test.ts` — 34 passed
- Headless browser against an isolated console, counting every `/api/v1/operations/catalog` request: 2 on load (Operations page + command band) and still 2 after cycling through four Operations (`GPT-5.6-Sol-Fast` -> `K3-1M` -> `Claude (Gateway)` -> `GPT-5.6-Sol-Fast`). No page errors.

## Notes
- Measurement correction for the record: "a normal startup with an active modeled Operation" does not reproduce as stated, because the active Operation is not persisted — a fresh load has no active Operation, so the before-fix bundle also issued exactly 2 requests. The duplicate needs the narrow window where the active coordinate changes before the first catalog response lands. The fix is still worth it: it closes that window and additionally drops the refetch when switching to a coordinate the index already knows.

sbluemin:
@codex Thanks — the ownership-scoping finding is deferred, with reasoning below. No further changes on this PR.

## Deferred
- P2 scope model labels to their catalog owner — `runtime/fleet-console/core/client/src/components/command-band-guards.ts:49`. The concern is real as a design axis, but it has no reachable path today and it is not this PR's scope:
  - `terminal`'s `claude-gateway` kind is the only kind in this repository that declares launch variants; no other built-in plugin and no example declares `variants` at all, so no coordinate collision exists to mislabel anything.
  - The repository already relies on the same premise elsewhere: `core/client/src/quick-launch.ts:findVariantLaunchKind` picks the first kind that declares variants. Scoping only the command band's index would leave two surfaces resolving the same coordinate under different contracts.
  - When an external plugin does start exposing launch variants, Quick Launch and the command band have to adopt an ownership model together — that is a separate decision about coordinate namespacing across installed plugins, not a change to make blind inside this PR.
  - Failure mode meanwhile stays bounded: an unknown coordinate falls back to the CLI label rather than breaking the surface.

## Passes
- Pass 1 (StrictMode ledger) — fixed by removing the ledger.
- Pass 2 (refetch on resolve) — fixed by depending on the coordinate itself plus a known-coordinate guard; the "every normal startup" framing did not reproduce and that correction is recorded above.
- Pass 3 — no FIX-class finding. Proceeding to merge.

PR #671, opened 2026-08-13
fix(quota): accept the six-card provider order after Grok

sbluemin:
## Summary

- Grok CLI (#668) added a sixth quota card (`xai`) on the client, but `POST /order` still required the pre-Grok five-id permutation and rejected every drag or arrow-key reorder.
- The panel then re-synced from summary, so the optimistic reorder snapped back and looked like the drop never applied.
- Share one provider-order list between the panel and the order route so a new card cannot leave the two sides on different contracts.

## Test Plan

- [x] `pnpm --filter @fleet-plugins/quota typecheck`
- [x] `pnpm --filter @fleet-plugins/quota test` (covers the six-card permutation the panel posts after a drag)
- [ ] Drag a quota provider card past another and confirm the new order stays after the drop and after a panel refresh


sbluemin:
@codex Please review this PR.

PR #695, opened 2026-08-14
fix(fleet-console): include the window caption in Tactical row stride

sbluemin:
## Summary
- Tactical grid/rows now include the 28px attached window caption in the row stride so stacked captions stay in their cells.
- Body and PTY geometry stay unchanged; the columns layout remains a single row.
- The shared caption constant, design-contract pins, and the unreleased caption note stay aligned.

Changelog-Amend: window-caption-b.md

## Test Plan
- [ ] Open Tactical with a multi-row grid and confirm each caption sits in its own cell without overlapping the row above
- [ ] Switch Tactical to rows and columns; columns stay a single full-height row, rows keep caption clearance
- [ ] Confirm War Room and maximize still shift the slot down and keep PTY size
- [ ] `pnpm exec vitest run tests/canvas-store.test.ts tests/instrument-design-contract.test.ts tests/operation-frame-identity.test.ts tests/canvas-minimap-formation.test.ts`

### DIFF
--- a/.changelog.d/window-caption-b.md
+++ b/.changelog.d/window-caption-b.md
@@ -4,5 +4,5 @@ branch: window-caption-b
 
 ### fleet-console
 #### Changed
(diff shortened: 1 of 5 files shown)


sbluemin:
@codex Please review this PR.

PR #696, opened 2026-08-15
fix(ai-gateway): page Cursor read tools directly

sbluemin:
## Summary

- Expose the caller `Read` schema directly to Cursor models so they can use bounded `offset` and `limit` paging.
- Keep `Grep` and shell native redirects, while rejecting native `Read` on the same Run with the advertised caller alias.
- Prevent repeated whole-file replay that measured 52 repeated reads and 10 compactions before the change, versus 1 repeated read and 2 compactions after it.

## Type of Change

- [ ] feat — new feature or carrier capability
- [x] fix — bug fix
- [ ] docs — documentation only
- [ ] refactor — no behavior change
- [ ] chore — build, tooling, deps
- [ ] BREAKING CHANGE

## Scope

- `packages/core-ai-gateway/src/cursor/native/`
- Cursor adapter and live Run bridge tests
- `fleet-cli` and `fleet-console` release notes

## Test Plan

- [x] `pnpm --filter @dotobokuri/core-ai-gateway test` (30 files, 662 tests)
- [x] `pnpm --filter @dotobokuri/core-ai-gateway typecheck`
- [x] `pnpm --filter @dotobokuri/core-ai-gateway build`
- [x] `git diff --check`
- [x] Run identical five-turn Cursor Composer 2.5 sessions before and after the change

## Changelog

- Let Cursor models page through caller-owned reads instead of repeatedly replaying whole files through failed native reads.
- Preserve caller permissions, fail-closed read semantics, and native search/shell redirects.

## Notes for Reviewers

The live comparison held the model, prompts, caller catalog, permissions, Theater, source revision, and logging constant. Exact duplicate Read output fell from 3,781,106 to 5,452 characters, workload replay bytes fell 27.4%, and automatic compactions fell from 10 to 2. Baseline turn 4 was queued while turn 3 was still running, so lifecycle-count deltas are not attributed solely to this patch; the exact Read duplication and direct paging evidence establish the tool-path effect.

### DIFF
--- a/.changelog.d/cursor-composer-compaction.md
+++ b/.changelog.d/cursor-composer-compaction.md
@@ -1,0 +1,13 @@
+---
+branch: cursor-composer-compaction
+---
+
+### fleet-cli
+#### Fixed
+- Let Cursor models page through the caller Read tool instead of repeatedly replaying whole files through failed native reads, reducing context growth and compaction during long gateway sessions.
+  ko: Cursor 모델이 실패한 네이티브 읽기로 전체 파일을 반복 재전송하는 대신 호출자 Read 도구로 페이지를 나눠 읽게 하여 장기 게이트웨이 세션의 컨텍스트 증가와 컴팩션을 줄였습니다.
+
+### fleet-console
+#### Fixed
+- Let Cursor models page through the caller Read tool instead of repeatedly replaying whole files through failed native reads, reducing context growth and compaction during long gateway Operations.
+  ko: Cursor 모델이 실패한 네이티브 읽기로 전체 파일을 반복 재전송하는 대신 호출자 Read 도구로 페이지를 나눠 읽게 하여 장기 게이트웨이 Operation의 컨텍스트 증가와 컴팩션을 줄였습니다.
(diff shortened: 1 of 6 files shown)


sbluemin:

## Addressed
- P2 parked native Read race — `packages/core-ai-gateway/src/cursor/native/adapter.ts` — restored native Read recognition only in the parked-frame classifier so the frame is deferred, then replayed into the attached segment where the direct caller Read policy response is written.
- Added a live Run regression test that reproduces the pre-fix bridge disposal and verifies deferred replay, same-Run completion, and the `cc_read_*` policy error.

## Validation
- Focused reproduction before fix — failed with `stream.closed === true`.
- Focused regression after fix — passed.
- `pnpm --filter @dotobokuri/core-ai-gateway test` — 30 files, 663 tests passed.
- `pnpm --filter @dotobokuri/core-ai-gateway typecheck` — passed.
- `pnpm --filter @dotobokuri/core-ai-gateway build` — passed.
- `node scripts/compile-changelog-fragments.mjs --check` — passed.
- `git diff --check` — passed.

## Notes
- Attached native Read execution remains excluded from caller-tool redirection and still returns the direct caller Read alias policy error.
- Grep and shell redirects and the caller permission boundary are unchanged.

PR #732, opened 2026-08-15
feat(fleet-console): add AI Gateway compact timing

sbluemin:
## Summary
- Settings > AI Gateway에 Compact timing(Auto / Early / Late / Custom)을 추가한다. Auto는 출하된 16k reserve, Early/Late/Custom은 각 모델 카탈로그 창의 88/97/70-99%.
- `compactCeiling`은 models PUT과 분리된 store write로 저장하고, 주입은 usage projection에만 한다. 413 가드·`[1m]`·catalog 창은 그대로다.
- Preview model은 절대 토큰만 보여 주고 저장하지 않는다.

## Test Plan
- [x] `tsc --noEmit` — `@dotobokuri/core-ai-gateway`, `@fleet-plugins/terminal`, `@dotobokuri/fleet-console`
- [x] gateway vitest (686) + settings-routes / settings-api (57) + compact-timing-track (3)
- [x] 격리 Console에서 Custom 트랙 드래그 확인 (70-99 축, 놓을 때만 저장)


sbluemin:

## Addressed
- P2 Keep compact timing stable for running sessions — `packages/core-ai-gateway/src/gateway-router/router.ts:343`. The gateway rereads `compactCeiling` on every `/v1/messages` call (same live-store path as the wire log). Snapshotting at a session boundary is out of scope: this package has no session identity. The settings copy now matches that lifetime — a change applies to the next gateway request.

## Validation
- `git diff aa014b3f..HEAD` is i18n only (en/ko help + foot).
- No runtime change.

## Notes
- Did not snapshot the policy per Operation. That would be a new durable session field across Console and the thin CLI gateway.

sbluemin:

## Addressed
- P2 Block the custom track while a save is pending — `runtime/fleet-plugins/terminal/client/agent/index.tsx`. `onPointerDown` now returns when `saving` is set, matching the other controls. The hidden range input was already disabled.

## Validation
- Tracked change is the `saving` guard on pointer-down and the live-track class.

## Notes
- First-pass P2 (session snapshot) stays declined for runtime snapshotting. Copy already matches live reread; the re-anchored comment is the same finding.

PR #733, opened 2026-08-15
feat(fleet-console): mention Quaker aides from Quick Launch

sbluemin:
## Summary

- Quick Launch `@` gains a second destination class: plugin-contributed non-Operation targets, through a new SDK pair — `mentionTargets()` and `messageMentionTarget(targetId, text)`. A plugin must declare both or its targets never reach the deck, matching the existing `messageOperation` / `messageableOperationTypes` pairing. `mentionTargets` is a function, not a static array, so the roster is re-read every time the deck opens.
- Scuttlebutt contributes the Quaker aides that are on duty. Sending moors that aide and the answer lands in its own speech bubble — the arrival-bubble envelope (360px, wrapping, dismiss, position tracking) wearing the chirp's brass, since the arrival `positive` is a state channel and an answer is not a state. "Open the full answer" hands over to the existing chat card. Aides that are off do not appear at all, so a default install shows no aide category; discovery stays with the settings section.
- Renames the mascots from admirals to aides in user-visible copy and in the system prompt's own-rank sentence, freeing "Admiral" to name only the host agent. Internal identifiers and the serialized `admiral` request key are deliberately unchanged.

## Test Plan

- [x] `tsc --noEmit` for `runtime/fleet-console` (covers the plugins)
- [x] `vitest run` in `runtime/fleet-plugins/scuttlebutt` — 17 files / 130 tests
- [x] `vitest run` for the touched console suites (quick-launch, quick-launch-layout, instrument-design-contract, i18n, effort-track, operation-search) — 6 files / 242 tests
- [x] Headless real browser on the built bundle: only the on-duty aide is listed; arrow keys cross the category boundary; the tag reads `부관 · 토리` with a separate `웹 전용` note; sending closes the composer and moors the aide (3px over 5s); the answer renders at 360px; "전문 보기" opens the chat card with the moor switch back to false; Escape dismisses; zero page errors
- [x] Adversarial review of the change (4 dimensions, 2 refuters per finding) — five confirmed defects fixed and re-verified live: a mention to a busy aide silently ate the draft; delivery awaited the whole answer so an unconnected SSE left the submit lock on forever; a second aide overwrote the single answer slot; turning an answering aide off left it moored; the bubble re-announced the whole answer on every streamed chunk
- [x] Rebased onto `canary` after #726–#729; both conflicts (the composer imports and the placeholder expression) integrated both sides, then revalidated and re-checked live


sbluemin:

## Addressed

- **P2 — overlapping concurrent bubbles** (`answer-bubble.tsx:55`) — **fixed** in `9f802494`. Reproduced first: with reduced motion emulated, `parkBirds` seats the two aides **92px apart** (measured x=1208 / x=1300, same row), and the two 360px bubbles overlapped **268 x 123px = 32,964px2**, covering ~74% of the first answer including its dismiss control. A bubble now steps past every earlier bubble it actually overlaps horizontally, using their measured heights; bubbles far apart keep their place, and an unpositioned first-frame sibling is skipped. Re-measured after the fix: overlap area **0**.

## Validation

- `tsc --noEmit` for `runtime/fleet-console` (covers the plugins) — pass
- `vitest run` in `runtime/fleet-plugins/scuttlebutt` — 18 files / 134 tests pass, including a new `client-answer-lane.test.ts` that pins the lane arithmetic (no move for the first bubble, steps past each overlapping earlier one, no move when horizontally clear, skips an unplaced sibling)
- Headless real browser on the rebuilt bundle with reduced motion emulated — two concurrent answers now read in separate lanes, both dismiss controls reachable, zero page errors

## Notes

- **P2 — "preserve an undismissed answer before accepting another"** (`flock.tsx:178`) — **declined.** Showing only the latest exchange is the shipped design of this mascot chat, not something this PR introduced: `chat-store.ts:41` states it as a contract ("카드는 현재 브라우저 메모리의 마지막 질문과 그에 대한 답만 보여준다"), and `chat-card.tsx:93` already renders `currentExchange(state)`. Asking an aide a second question from its card has always replaced the visible exchange; the mention path reaches the same model through a different door rather than creating a new loss.

  The proposed remedy would also make the deck **stricter than the card** - you could ask again from the card but not from Quick Launch - which narrows supported behavior. Addressing an aide again is the intended behavior, and the single-visible-exchange trade-off is recorded in this PR's frozen context. Reconsidering it would be a change to the Quaker chat model itself, which is outside this PR.

  The related loss this PR *did* create was a different one and is fixed in `e3e9cf3b`: a mention to an aide that was still answering used to resolve as delivered while `ChatSession.ask` dropped it, so the composer cleared a draft that never went anywhere. That now rejects with `destination_busy` and keeps the draft.

sbluemin:

## Addressed

- **P2 — anchor deltas ignored when stacking** (`answer-bubble.tsx:63`) — **fixed** in `878a713f`. Reproduced first, with normal motion: dragged two aides to a horizontally overlapping, vertically offset arrangement (**47px apart**), asked both, and measured **75 x 77px = 5,775px2** of residual overlap. The summed-height lane was indeed blind to the anchor delta.

  The lane now tests the rectangle the bubble is about to occupy against each earlier bubble it overlaps horizontally, and retreats exactly past that bubble's outer edge — `anchorTop - rect.top` opening upward, `rect.bottom - anchorBottom` opening downward — keeping the largest retreat any earlier sibling demanded. Same-row aides reduce to the previous arithmetic.

## Validation

- Re-measured the exact failing arrangement on the rebuilt bundle: overlap **0**, with the two bubbles separated by precisely the 8px gap
- Regression-checked the reduced-motion case from the previous pass: aides parked **92px** apart, overlap still **0**
- `vitest run` in `runtime/fleet-plugins/scuttlebutt` — 18 files / 137 tests pass. `client-answer-lane.test.ts` now pins the anchor-delta case directly: the same pair of bubbles yields lane `128` when the mascots share a row and `175` when the later mascot sits 47px lower, plus the downward-opening case, the horizontally-clear case, the already-vertically-clear case, and the unplaced-sibling case
- `tsc --noEmit` for `runtime/fleet-console` — pass; `pnpm --filter @dotobokuri/fleet-console build` — pass
- Zero page errors across both live runs

## Notes

- The previous pass's decline stands unchanged: showing only the latest exchange is the shipped Quaker chat model (`chat-store.ts:41`, `chat-card.tsx:93`), not something this PR introduced, and gating the deck on an undismissed bubble would make it stricter than the card.

sbluemin:
@codex The review feedback has been addressed.

## Addressed

- **P2 — rescan earlier bubbles after each lane retreat** (`answer-bubble.tsx:208`) — **fixed** in `9df466bb`. Your counterexample reproduced exactly. Fed straight into `laneOffset`: earlier bands `[100,200]` and `[220,320]`, candidate starting at `[250,350]`, `gap` 8 — the pass skipped the first band, cleared the second, retreated to `[112,212]`, and `overlapsA` came back **true**. Because every frame restarts at `lane = 0`, it persisted, exactly as you described.

  The scan now repeats until a pass moves nothing, bounded by the number of blocking bubbles (three aides, so a handful of turns at most). Blockers are collected once per placement rather than re-queried inside the loop.

## Validation

- Unit: `client-answer-lane.test.ts` pins your numbers directly — the three-band case now ends with the candidate's bottom edge at or above `100`, plus the existing same-row, anchor-delta, downward-opening, horizontally-clear, vertically-clear and unplaced-sibling cases. `vitest run` in `runtime/fleet-plugins/scuttlebutt` — 18 files / **138 tests** pass
- Live: all three aides dragged into a horizontally overlapping, vertically staggered arrangement and asked concurrently — **all three pairwise overlaps 0**, every answer and control readable, zero page errors
- `tsc --noEmit` for `runtime/fleet-console` — pass; `pnpm --filter @dotobokuri/fleet-console build` — pass

## Notes

- This closes the review loop at the third pass, per the workflow's stop rule. Each pass produced a reproduced, fixed defect and the fixes stayed inside one function and its test; I am not opening a fourth round for further placement refinement. The bubble placement is now: anchor above or below by the mascot's screen half, clamp horizontally to the viewport, and retreat past horizontally-overlapping earlier bubbles until stable.
- The pass-1 decline stands: showing only the latest exchange is the shipped Quaker chat model (`chat-store.ts:41`, `chat-card.tsx:93`), not something this PR introduced, and gating the deck on an undismissed bubble would make it stricter than the card that has always allowed a follow-up question.

PR #737, opened 2026-08-15
feat(fleet-console): stand the real panel in every War Room deck tile

sbluemin:
## Summary
- War Room 덱의 한 칸이 이제 카드가 아니라 **자리**입니다. 칸은 빈 mount 노드를 내놓고, 캔버스가 그 Operation의 실제 `OperationFrame`을 그 자리로 portal합니다 — 패널 소유권은 끝까지 캔버스에 있고 바뀌는 것은 DOM 상의 부모뿐이라 상태·이벤트·body pool 배선이 그대로 유지됩니다.
- 덱 칸에 선 프레임은 캔버스 좌표를 버리고 칸의 상자를 그대로 입습니다. 터미널이 그 상자를 재므로 **PTY가 밀도에 따라 리사이즈되고 셸이 그 칸에 맞춰 다시 줄바꿈**합니다. 종전의 축소 프리뷰 산술(`resolveTriagePreviewFit`·`MinScale`·`ContentHeight`)과 카드 얼굴은 함께 폐기했습니다.
- War Room 고유 레이아웃은 그대로입니다: Theater 밴드, 덱 줌(0.35–2.0), Quick-Look 확대, 밀도 변형(칸↔점), 임계 아래 지도 점. 지도 Quick-Look은 별도 얼굴 대신 같은 패널을 판 위로 올립니다.
- 덱이 바닥에 예약하던 74px 띠를 회수했습니다 — 실측 결과 그 자리에 닿는 것은 캔버스 배경뿐이었고, 그만큼 칸 높이와 한 행의 수용량을 깎고 있었습니다.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console test` — 2149 passed / 24 skipped
- [x] 격리 Console 헤디드 실측(1680×1010): 칸 302×210에 실물 패널·터미널 300×177, 밀도 1.6×에서 425×303으로 PTY 재줄바꿈, 0.4×에서 지도 점 14px 유지, 점 hover 시 같은 패널이 판 위로, 칸 클릭 → 무대 승격, 최소화·닫기 2단 확정, 덱 하단 여백 75px → 27px


sbluemin:

## Addressed
- **P2 — tile signals never matched** (`components.css:3420,3562,3566,3570` + reduced-motion block): confirmed by building the real DOM (`cell > mount > panel`) and asserting `Element.matches` — every direct-child selector returned `false`. All four signal selectors now pass through `.canvas-triage-deck-mount`. Live check on the isolated Console: `is-fresh` -> `triage-fresh-pulse`, `is-arriving` -> `triage-deck-arrival-pulse`, `is-landed` -> `triage-deck-landed-flash`, idle -> `none`, and the map emphasis repaints the panel border.
- **P2 — unidentified blank tile** (`triage-watch-deck.tsx:1078`): the mount now carries `data-fallback-title={operation.title}` and `.canvas-triage-deck-mount:empty::after` draws it while no panel can populate the slot. The promotion surface still covers the tile, so a click promotes the same Operation. Live check: emptying a mount renders `content: "Shell"` inside a dashed placeholder.
- You were right that the contract could not catch this: it only asserted the selector text existed. The new `triage-store` test extracts each signal selector from the stylesheet and runs `matches()` against the portaled panel, so a selector that paints nothing now fails.

## Validation
- `pnpm --filter @dotobokuri/fleet-console test` — 2150 passed / 24 skipped (the new regression test fails against the previous CSS)
- Headless Console at 1680x1010 — signals, fallback, and map emphasis measured as above

## Notes
- No declines this pass.

sbluemin:

## Addressed
- **P2 — deck panel bodies reachable by keyboard** (`canvas.tsx:1301` / `operation-frame.tsx`): confirmed — the promotion surface only intercepts the pointer, so tabbing walked into the live plugin body before reaching it. The deck tile's `.canvas-operation-terminal` is now `inert`, so the tab order is caption controls then promotion surface and stops there. The caption stays reachable on purpose: minimize and close remain keyboard-operable, which is why the frame marks the body rather than the whole article, and a frame standing on the canvas is unaffected.

## Validation
- `pnpm --filter @dotobokuri/fleet-console test` — 2152 passed / 24 skipped
- New behavioural test in `operation-frame-body-activation.test.ts` asserts the deck tile's body carries `inert`, the caption does not, and the canvas frame keeps its body live; reverting the fix makes it fail.

## Notes
- No declines this pass.

sbluemin:

## Addressed
- **P2 — Quick-Look dropped over the portaled caption** and **P2 — caption context menus bypassed the deck**: both reproduced on the live Console before editing. Moving the pointer from the promotion surface onto the caption took the magnified tile from 1 to 0, and a `contextmenu` dispatched on the caption left `defaultPrevented === false` while the same event on the promotion surface returned `true`.
- Root cause is one thing, as you noted: the portaled frame is the canvas's child in the React tree, so the cell's synthetic handlers never observe its caption. The plate now delegates `pointerover`, `pointerout`, and `contextmenu` as **native** listeners on the deck grid, which follow DOM bubbling — the same reason `OperationFrame` already listens natively for `pointerdown` on its transplanted body. Enter/leave is decided by comparing the `relatedTarget`'s cell, so moving within a tile (body -> caption -> controls) is not a leave.
- Measured after the change: Quick-Look reads 1 on the body, 1 on the caption, 0 after leaving the cell; caption, body, and promotion-surface right-clicks all route to that Operation with the tile's promotion surface as the focus-return target; bare band space still opens the Theater menu.

## Validation
- `pnpm --filter @dotobokuri/fleet-console test` — 2153 passed / 24 skipped
- New regression test builds the portaled structure (cell > mount > panel > caption) and asserts the caption's context menu reaches the Operation; removing the delegated listener makes it fail.

## Notes
- No declines this pass.

sbluemin:

## Addressed
- **P2 — companion frames portaled into the tile** (`canvas.tsx:1389`): confirmed by tracing the path rather than assuming it — `pages/operations.tsx` opens a companion for whatever Operation is active, with no War Room guard, and a deck panel's caption sets that active Operation through the frame's own `onPointerDown`. Since `renderPluginOperation` returns the frame and its `CompanionFrame`s as one fragment, the whole companion layout went into the tile mount. The deck slot now excludes a companion-open panel, so it renders on the canvas as before and the tile keeps only its name.

## Validation
- `pnpm --filter @dotobokuri/fleet-console test` — 2153 passed / 24 skipped
- Design contract pins the slot condition so a later edit cannot silently drop the companion guard.

## Notes
- No declines this pass.

sbluemin:
@codex Thanks — this one I am deferring rather than fixing, with evidence.

## Deferred (not a regression from this PR)
- **P2 — companion's primary frame hidden in War Room** (`canvas.tsx:851`): the hidden primary frame is `canary`'s behaviour, not something this PR introduces. On `canary`, both branches read only the stage:
  - `focusLayerHidden = triageActive ? !operationTriageStage : ...` (canary canvas.tsx:871-873)
  - `minimized: triageActive ? !operationTriageStage : ...` (canary canvas.tsx:926)

  A companion-open, non-staged Operation is therefore hidden there too. This PR's branches read `!operationTriageStage && !deckSlot`, and the companion guard makes `deckSlot` null for exactly that Operation — so the resulting state is identical to `canary`. What the previous pass fixed was the new problem this PR had introduced (companion frames portaled into the tile); it did not change whether the primary frame is visible.

  Making the companion's primary frame visible in War Room is a product-behaviour question about a path that predates this change, so it belongs in its own change rather than here. The scope frozen for this PR is: real panels in the deck tiles, PTY resizing with density, War Room's layout and density axis preserved, and the reclaimed bottom band.

## State
- `pnpm --filter @dotobokuri/fleet-console test` — 2153 passed / 24 skipped; typecheck and build pass.
- Five findings across four passes were reproduced and fixed; this is the first I am deferring, on the evidence above.

PR #757, opened 2026-08-16
feat(fleet-console): answer the agent inside the chat view

sbluemin:
## Summary

- 채팅 뷰가 `AskUserQuestion`과 `ExitPlanMode`를 받아, 모델이 멈춰 물은 자리에 카드를 세우고 그 답으로 턴을 잇습니다. 두 도구는 SDK가 `canUseTool` 콜백을 받을 때만 자식에게 실리므로(실측 0.3.212: 29 → 32), `ClaudeGatewayTurn` allowlist에 그 키 하나를 엽니다. 이 콜백은 권한 게이트가 아니라 답을 돌려주는 통로이고, `permissionMode`는 그대로 `bypassPermissions`입니다.
- 답이 자식에게 닿는 통로는 권한 응답 하나입니다: 질문은 `updatedInput.answers`, 계획 승인은 allow 자체, 수정 요청은 deny의 message(모델이 계획을 고쳐 다시 냅니다). 카드의 입력은 대화 입력창이 아니라 그 질문에만 사는 답변 칸이라, 지시가 Quick Launch로만 들어온다는 결정은 그대로입니다.
- 대기는 만료되지 않으므로(SDK는 권한 요청에 park deadline을 두지 않습니다) 상태로 표면화합니다: `chatAwaiting`을 사이드바 칩·War Room 타일·맵 도트가 `awaiting`으로 읽고, 나가는 문(건너뛰기·수정 요청·턴 중단·dispose)이 모두 대기를 정리합니다.
- `console-handoff` 스킬을 함께 넣습니다 — 작업 브랜치 빌드로 격리 Console을 띄우고 시나리오를 시드해 사람에게 URL로 넘기는 절차입니다.

## Test Plan

- [x] `pnpm exec vitest run tests/ ../fleet-plugins/terminal/tests/` — 2210 passed / 24 skipped
- [x] `pnpm exec vitest run` (core-agent) — 122 passed (신규 계약 2건 포함)
- [x] `npx tsc --noEmit -p runtime/fleet-console/tsconfig.json`
- [x] 격리 Console + `claude-gateway--cursor--auto` 실측 왕복: 카드 렌더 → 브라우저에서 선택지 클릭 → `settled(answered)` → 모델이 그 답을 채택
- [x] 자유 입력(목록 밖 문자열) 왕복: "Redis, actually" → 모델이 그대로 채택
- [x] `awaiting` 축: 대기 중 사이드바 칩 `is-awaiting`, 답변 후 `is-idle`
- [ ] `ExitPlanMode` 실왕복은 e2e 미검증 — cursor auto가 그 도구 대신 텍스트로 계획을 끝냈습니다(모델 행동). 서버 경로는 단위 테스트로 고정했고, 그 도구가 콜백으로 온다는 사실은 native Claude 프로브로 실측했습니다.

## Notes

- 계획 파일 되쓰기(`planFilePath`를 세션 밖으로 옮기는 일)는 이번 범위에서 뺐습니다. 격리 config dir이 지워질 때 계획이 함께 사라지는지를 확인하지 못했고(인증 때문에 격리 dir 프로브가 막혔습니다), 미검증 전제 위에 코드를 얹지 않았습니다.


sbluemin:

## Addressed

- **P2 — answer key** (`chat-session.ts:272` → `chat-events.ts`): confirmed. The key came from the display text, which passes `trim()` and a 400-char cap, so a padded or long question produced a key `AskUserQuestion` could not match — the route answered 200 and the card folded while the model resumed with no answer. `agentChatAskFromToolInput` now returns `{ event, answerKeys }`: the event stays bounded for the browser, and `answerKeys` carries the untouched question strings the tool sent. The session keys `updatedInput.answers` by those. Keys are pushed in step with the questions that actually shipped, so a skipped question cannot shift the index. Regression tests cover both the parser and a full parked-answer round trip with a >400-char padded question.
- **P2 — single-select selected state** (`chat-view.tsx:667`): confirmed. `aria-pressed` was conditional on `multiSelect`, and the only selected styling keys off that attribute, so with several single-select questions in one ask the earlier answer was invisible until the last question completed. Every option now reports `aria-pressed`, which also lights the existing `[aria-pressed="true"]` styling. Kept to toggle-button semantics rather than converting the group to radios: that would change the keyboard contract for a control that is one tap in the common single-question case.
- **P1 — design contract** (`chat.css:413`): confirmed against the repository rule that a grammar change moves with the contract. Added `pins the chat question-card grammar` to `instrument-design-contract.test.ts` — aurora carries the wait with no other signal token in the card block, brass appears only on hover/focus and the single primary action, the free-text field is dashed until focus, the settled line keeps positive/coral, and the wait pulse is listed in the reduced-motion cutoff.

## Validation

- `pnpm exec vitest run` (runtime/fleet-plugins/terminal) — 79 files, 779 passed
- `pnpm exec vitest run` (runtime/fleet-console) — 181 files, 2220 passed / 24 skipped
- `pnpm exec vitest run` (packages/core-agent) — 122 passed
- `npx tsc --noEmit -p runtime/fleet-console/tsconfig.json` — clean
- `pnpm --filter @dotobokuri/core-agent build` + `--filter @dotobokuri/fleet-console build` — clean

## Notes

- The branch was rebased onto canary before this pass; canary had reworked the same chat view for scroll-follow, and both changes were verified to survive together.
- Still out of scope by decision: writing an approved plan out of the session's temporary config dir. The premise (that `planFilePath` lives under a directory disposal deletes) is unverified, and `ExitPlanMode` was never emitted by the pinned model during live testing, so that path stays measured-first rather than coded on an assumption.

sbluemin:

## Addressed

- **P2 — option label** (`chat-events.ts:395`): confirmed, and it is the same class of defect as the question key from the last pass. The label is answer data, not display data: the card submits the label it renders, so an ellipsized label sent the model a value it never offered. Rather than carrying a second copy, the cap is gone — a label is now passed through untouched and the card wraps it (`overflow-wrap: anywhere`). `header` keeps its cap because it is only ever displayed. Test: a 200+ character label survives the parser unchanged.
- **P2 — already-aborted signal** (`chat-session.ts:370`): confirmed. `AbortSignal` does not replay a past abort, so a listener registered after the fact never fires. One correction to the finding: the promise was not parked indefinitely — `runTurn`'s `finally` calls `abandonAsks`, which settles every parked question when the stream ends, and that is the same net that covers a lost abort today. What was real is the window before that: the panel showed a question for a turn that had already ended and the sidebar reported the operation as awaiting. The callback now checks `signal.aborted` before raising anything and denies straight away. Test: a pre-aborted signal returns the denial, publishes no card, and reports no awaiting.

## Validation

- `pnpm exec vitest run` (runtime/fleet-plugins/terminal) — 79 files, 781 passed
- `pnpm exec vitest run` (runtime/fleet-console) — 181 files, 2220 passed / 24 skipped
- `npx tsc --noEmit -p runtime/fleet-console/tsconfig.json` — clean
- `pnpm --filter @dotobokuri/core-agent build` + `--filter @dotobokuri/fleet-console build` — clean

## Notes

- Both passes have now landed on the same root shape: anything the user can choose is protocol data and must reach the tool byte-for-byte, while display bounds belong to CSS. The question key, and now the option label, are the two places that carried it.

sbluemin:
@codex Addressed one and declined one. Please re-review.

## Fixed

- **P2 — unseen plan approved** (`chat-events.ts:370`): confirmed, and this one differs in kind from the label findings. Approval means agreeing to what was shown; a truncated plan broke that equality, because the card rendered a prefix while `pending.input` handed the SDK the whole document. The cap now matches the text event's (60k), which puts any ordinary plan inside it, and a plan that still exceeds it carries `truncated: true`. The card closes the approve button for that case and points at the change-request path, which already exists, to ask for a shorter plan. Approving something unseen is now structurally impossible rather than merely unlikely.

## Declined

- **P2 — option label whitespace** (`chat-events.ts:392`): declined, with the same evidence that drove the two fixes. The value in `updatedInput.answers` is not a matching key — it is free text. Measured on this branch: a string the tool never offered ("Redis, actually" against options JSON / Plain text) was accepted verbatim and the model adopted it. So the previous two findings were real for reasons whitespace does not share: the question text **is** the key the tool matches on, and an ellipsized label **lost content**. Leading or trailing spaces change neither what the user chose nor what the model understands, and preserving them would split the label into display and wire copies again — including inside the comma-joined multi-select value, where a raw label would carry its padding into the joined string. The cost lands on the user's answer; the benefit is byte-equality for its own sake.

If there is a case where the tool compares an answer value rather than treating it as free text, that would overturn this and I will take the fix — but the measurement on this wire says otherwise.

## Validation

- `pnpm exec vitest run` (runtime/fleet-plugins/terminal) — 79 files, 782 passed
- `pnpm exec vitest run` (runtime/fleet-console) — 181 files, 2220 passed / 24 skipped
- `npx tsc --noEmit -p runtime/fleet-console/tsconfig.json` — clean
- `pnpm --filter @dotobokuri/fleet-console build` — clean

sbluemin:
@codex Fixed. This closes the review loop from my side — see the note at the end.

## Fixed

- **P2 — approval guard was client-only** (`chat-session.ts:247`): confirmed, and it is a defect in my own previous fix rather than a new finding. I claimed approving something unseen was structurally impossible while the guard sat in `AskCard`, which makes it a courtesy of one rendering, not a rule: a tab that has not reloaded still draws the button, and its `{approve: true}` reached a session that held the whole plan with no memory that only a prefix had been shown. `PendingAsk` now carries the truncation and `answer()` refuses approval there. The refusal keeps the question parked — releasing it would take away the only place left to answer — and leaves the change-request path open, so asking for a shorter plan is still the way forward. The route answers 409 rather than 400: the request is well formed, the session simply cannot approve what it could not show. Test: a 60,050-character plan is refused on approval, stays parked, and still accepts a revision.

## Validation

- `pnpm exec vitest run` (runtime/fleet-plugins/terminal) — 79 files, 783 passed
- `pnpm exec vitest run` (runtime/fleet-console) — 181 files, 2220 passed / 24 skipped
- `npx tsc --noEmit -p runtime/fleet-console/tsconfig.json` — clean
- `pnpm --filter @dotobokuri/fleet-console build` — clean

## Closing the loop

Four passes, seven findings: five fixed, one declined with measurement, and this last one a correction to my own incomplete fix. Thank you — the answer-key and option-label findings were real defects that would have failed silently in production, and this one caught a guard I had put in the wrong layer.

I am stopping the review cycle here per this repository's workflow, which ends the loop on the pass that yields no further FIX-class finding and caps it at the third pass by default; this fourth pass was taken because it named a specific reproducible defect. Further suggestions on this surface are welcome as follow-up issues rather than blockers for this PR — particularly anything about the plan file's lifetime outside the session config dir, which is deliberately out of scope here and noted in the PR body.

PR #763, opened 2026-08-16
feat(fleet-console): open chat background work beside the conversation

sbluemin:
## Summary

- 작업 탭을 걷고, 백그라운드 작업이 **대화 옆에** 섭니다. 넓은 패널에서는 오른쪽 컬럼, 좁은 패널에서는 아래 서랍이고, 그 전환은 뷰포트가 아니라 **패널 자신의 폭**이 가릅니다(`chat.css` 최초의 `@container`). 하단 스트립이 유일한 문이며, 면이 열리면 스트립이 그 면의 머리로 옮겨 앉습니다 — 떠 있는 알약으로 남으면 열린 면의 첫 줄을 덮고 접는 문이 자란 것 밖에 섭니다(실측으로 확인해 고쳤습니다).
- **턴 중지**를 추가했습니다. 도는 턴만 끊고, 큐에 밀려 있던 턴까지 함께 버리며, 결말을 실패가 아니라 `stopped`로 세웁니다. 잡 하나만 멈추는 기능은 넣지 않았습니다 — `TaskStop`은 모델 도구이고 호스트 제어 경로가 SDK에 없습니다(0.3.212 확인).
- 잡 상세에 **서브에이전트 도구 발자국**(사이드체인 전사록에서 도구 호출만)과 **셸 출력 꼬리**를 붙였습니다. 둘 다 잡을 연 그때만, 파일 끝에서 경계 창만큼만 읽습니다.
- 접힌 도구 호출 줄의 시인성을 고쳤습니다. 쉬는 상태에서 눌리는 줄과 안 눌리는 줄의 차이가 9px `--hairline-strong` 글리프 하나뿐이었습니다.

`effort` 표시는 **의도적으로 제외**했습니다. `AgentInput`·`AgentOutput`·`workflow_agent` 행·`meta.json` 2,905건 어디에도 반환값이 없어, 넣으면 없는 값을 지어내거나 빈 열을 그리게 됩니다.

Changelog-Amend: chat-background-work.md

## Test Plan

- [x] `pnpm --filter @fleet-plugins/terminal test` — 852 passed (80 files)
- [x] `pnpm --filter @dotobokuri/fleet-console test` — 2220 passed, 24 skipped
- [x] `pnpm build` — 전체 통과
- [x] 경계 게이트 4종(`core-boundary`, `claude-agent-sdk-boundary`, `core-unified-agent-absent`, `localized-attributes`) 통과
- [x] `node scripts/compile-changelog-fragments.mjs --check` — 18 entries OK
- [x] 격리 콘솔 실 E2E(cursor-auto): 중지 버튼이 턴 중에만 서고 누르면 `is-stopped`로 닫히며 오류 줄이 서지 않음 · 952px에서 `row`/600px에서 `column`으로 전환(서랍이 전폭 600px) · 스트립이 실행 중/쉬는 두 형태로 관측 · 실제 백그라운드 셸 출력 `beat 1..7`이 꼬리로 렌더 · 집계 줄 꺾쇠 11px `--text-tertiary`·이름 `--text-secondary` 실측
- [x] 스크린샷으로 겹침 결함 1건 발견 후 수정 — 떠 있는 칩 줄이 접기 컨트롤을 덮어 `elementFromPoint`가 칩을 집던 것을 대화 면 안으로 옮겨 해결
- [x] 리뷰 3패스 6건 전건 수정(declined 0). 각 수정마다 수정 전 red를 확인했고, 그중 2건은 제 앞선 수정이 만든 결함이라 그렇게 기록했습니다.


sbluemin:

## Addressed
- **P1 bounded tail read** — `chat-session.ts:readShellTail` — the output file is now opened and read at an offset for the last 256 KB only, instead of `readFile`-ing it whole. When the window starts inside the file the first partial line is dropped (it can also be a split UTF-8 sequence) and `truncated` is set. Your reading matched a measurement already in this PR's record: a real background shell had produced a 438 KB file, and a backgrounded build or test run is orders of magnitude larger.
- **P2 tail cap direction** — `chat-events.ts:chatShellTailFromOutput` — the character cap now slices from the end before sanitizing. `capTo` keeps the head, so a command emitting long JSON lines showed the oldest of its last 200 lines under a label that said "showing the tail". Slicing before sanitizing is safe because every sanitizer step only shortens.
- **P2 detail retry after the notification** — `chat-view.tsx:useAgentChatJobDetail` — the effect now also depends on the fields the notification brings (`summary`, `durationMs`), so the request re-fires on the render where the outcome lands. This one was real and I had already seen half of it: in the first live run the shell detail read "no record found", and I attributed all of it to a wrong path. Fixing the path removed one cause; this race was the other, and it reproduces whenever the detail is open at the moment the job ends.

## Validation
- `pnpm --filter @fleet-plugins/terminal test` — 848 passed (80 files)
- `pnpm --filter @dotobokuri/fleet-console test` — 2220 passed, 24 skipped
- `pnpm build` — pass; `pnpm --filter @fleet-plugins/terminal typecheck` — pass
- `node scripts/compile-changelog-fragments.mjs --check` — 18 entries OK
- Each fix has a test verified to fail without it. The bounded read is pinned with a spy rather than by content, because the 200-line cut yields the same text whether the read was windowed or whole — content alone cannot tell the two apart.

## Notes
- No declines and nothing deferred; all three were reproduced before editing.

sbluemin:

## Addressed
- **P2 empty tail on an oversized line** — `chat-session.ts:readFileTail` — correct, and this one was created by the previous pass's fix: aligning to a line boundary dropped the whole buffer whenever the only newline sat at the very end of the window, or when there was none. Alignment is now skipped when it would leave nothing, and the split UTF-8 prefix is stripped at the byte level (leading `10xxxxxx` continuation bytes) rather than relying on the newline to carry it away. I considered rolling the previous fix back per "a fix that breeds the next finding was over-scoped", but rolling back restores an unbounded read, which is strictly worse — so this finishes that fix instead of widening it.
- **P1 bounded subagent transcript read** — `chat-session.ts:readJobDetail` — the transcript now goes through the same window. Its window is larger than the shell one (4 MB vs 256 KB) because a JSONL line carries a whole tool result: a measured sample held 126 KB in 19 lines, so a narrow window would shorten the trail for a reason the reader cannot see. That is a different lie from the 200-step cap, which the UI already reports.

Both readers now share one `readFileTail`.

## Validation
- `pnpm --filter @fleet-plugins/terminal test` — 850 passed (80 files)
- `pnpm --filter @dotobokuri/fleet-console test` — 2220 passed, 24 skipped
- `pnpm build` — pass
- Both fixes have tests verified to fail without them: removing the alignment guard empties the tail for a 512 KB single line, and restoring `readFile` for the transcript makes the head marker reappear and the `readFile` spy fire.

## Notes
- No declines and nothing deferred.

sbluemin:
@codex Addressed. This is the third pass, and per this repo's stop rule it is the last one I will act on unless a reproduced defect justifies more.

## Addressed
- **P2 retry after a status-only notification** — `chat-events.ts` / `chat-view.tsx` — you are right, and the evidence you cited is mine: `jobEndEvent` deliberately accepts a status-only notification and a test already covers that shape. My first-pass fix inferred arrival from `summary`/`durationMs`, which is exactly the inference that fails there. The job record now carries `ends`, a count of end reports; nothing renders it, it carries the arrival itself. The regression test was rewritten to the status-only shape and now fails against the previous implementation — I verified that by restoring the old dependencies and watching it go red.

## Validation
- `pnpm --filter @fleet-plugins/terminal test` — 852 passed (80 files)
- `pnpm --filter @dotobokuri/fleet-console test` — 2220 passed, 24 skipped
- `pnpm build` — pass
- Added a reducer test pinning that two identical status-only end reports still move the counter, since no other field can tell the second arrival apart.

## Notes
- Across three passes: 6 findings, 6 fixed, 0 declined, 0 deferred. Two of the six were defects my own earlier fixes created, which I recorded as such rather than as new scope.

PR #765, opened 2026-08-16
fix(fleet-console): drop Chat EventSource on parked bodies

sbluemin:
## Summary
- Chat Mode used to keep an EventSource open for every parked, minimized, hidden, or War Room deck-tile body. Those sockets filled Chromium's HTTP/1.1 per-host slots, so panel close and dormant Resume stayed pending while the visible chat stream kept flowing.
- The host now publishes `bodyLive` on the operation render context. Chat Mode opens `/chat-stream` only while the body is readable; an older host that omits the field still subscribes.
- Parked bodies keep the last journal in memory (`idle`, not `lost`) so returning to the panel does not flash a connection error before replay.

## Test Plan
- [x] `pnpm --filter @fleet-plugins/terminal exec vitest run client/agent/chat/chat-store.test.tsx client/agent/chat/chat-work-tab.test.tsx client/agent/chat/chat-scroll-anchor.test.tsx`
- [x] `pnpm --filter @dotobokuri/fleet-console exec vitest run tests/companion-availability.test.tsx tests/operations-boot-minimization.integration.test.ts`
- [x] `pnpm --filter @dotobokuri/fleet-console exec tsc -p tsconfig.json --noEmit`
- [x] `pnpm --filter @fleet-plugins/terminal exec tsc -p tsconfig.json --noEmit`
- [ ] With several Chat Mode operations open, minimize or leave some off-screen, then close another panel and click Resume on a dormant one — both should settle instead of hanging on "재개 중…".

### DIFF
--- a/.changelog.d/chat-parked-stream-off.md
+++ b/.changelog.d/chat-parked-stream-off.md
@@ -4,5 +4,5 @@ branch: chat-parked-stream-off
 
 ### fleet-console
 #### Fixed
-- Chat Mode no longer holds a live stream for a parked, minimized, hidden, or deck-tile panel. Only a body the user can read keeps its EventSource, so close and Resume requests are not starved behind off-screen chat subscriptions.
-  ko: 채팅 모드가 주차·최소화·숨김·덱 타일 패널에 라이브 스트림을 붙들지 않습니다. 사용자가 읽는 본문만 EventSource를 열어, 화면 밖 채팅 구독 때문에 닫기·재개 요청이 줄 서지 않습니다.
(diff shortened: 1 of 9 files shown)


sbluemin:

## Addressed
- P1 hydrate deck tiles before suppressing streams — `runtime/fleet-console/core/client/src/canvas/canvas.tsx`: War Room deck tiles stay `bodyLive`. Their bodies are painted (inert only). Parked, minimized, and hidden bodies still drop the EventSource.

## Validation
- `node scripts/compile-changelog-fragments.mjs --check` — OK
- `pnpm --filter @fleet-plugins/terminal exec vitest run client/agent/chat/chat-store.test.tsx client/agent/chat/chat-work-tab.test.tsx client/agent/chat/chat-scroll-anchor.test.tsx` — 15 passed
- `pnpm --filter @dotobokuri/fleet-console exec tsc -p tsconfig.json --noEmit` — pass

## Notes
- A one-shot replay snapshot for parked bodies was not added. The supported path that showed a blank journal was the deck tile; those tiles now stay subscribed.

PR #787, opened 2026-08-16
feat(fleet-console): keep the Chat Mode agent session alive across turns

sbluemin:
## Summary
- Chat Mode는 턴마다 SDK 실행을 새로 세우고 문자열 프롬프트를 넘겼습니다. vendor SDK는 그것을 single-turn으로 보고 **첫 `result`에 자식의 stdin을 닫으며**, 입력이 닫힌 자식은 Claude Code의 print wind-down 규율에 따라 **백그라운드 셸을 5초 유예 뒤 kill**합니다(에이전트·워크플로는 기다려 주지만 셸은 기다려 주지 않습니다). 즉 백그라운드 작업은 자기를 시작한 턴을 넘길 수 없었습니다.
- `core-agent`에 `startTurn`과 나란히 세션 표면을 추가했습니다. `openSession()`이 자식 하나를 호출자 수명 동안 붙들고, 스스로 끝나지 않는 입력 스트림으로 프롬프트를 밀어 넣습니다. Chat Mode는 Operation당 세션 하나를 열고 영속 리더로 읽으며, 턴 경계는 스트림의 끝이 아니라 `result` 메시지가 정합니다.
- 턴 중지는 자식을 죽이는 대신 `interrupt()`로 끊고, `stopTask()`로 백그라운드 작업 하나만 멈추는 문을 열었습니다. 자식이 사라지면 살아 있던 잡을 `stopped`로 닫아, 오지 않을 결말을 화면이 기다리지 않게 합니다.

## Test Plan
- [x] `pnpm typecheck` — 전 워크스페이스 통과
- [x] `pnpm build` — 전 워크스페이스 통과
- [x] `pnpm test` — 전 워크스페이스 통과 (terminal 플러그인 892건 포함)
- [x] 실 SDK 실측: 문자열 프롬프트는 `result` +5.17s에 셸이 `killed`, 스트리밍 세션은 20초 관찰 창 내내 생존, `stopTask`는 3ms에 결말 도착
- [x] 격리 Console 실측: 백그라운드 셸이 턴 종료 후 15초간 생존, `chat-job-stop` 라우트가 6ms에 잡을 멈추고 `job-end status=stopped` 도착, 발급된 적 없는 좌표는 409
- [ ] 잡 상세의 중단 버튼 렌더는 브라우저에서 확인하지 못했습니다 — 헤드리스에서 Chat 패널이 WebSocket을 붙이지 않아(이 변경 밖의 경로) UI 표면까지 닿지 못했습니다


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed
- P2 `chat-session.ts:1030` — `retireSession` now calls `session.close()` before clearing the reference, so the SDK instance frees its single active slot. Confirmed the diagnosis: without it the next `openSession()` is rejected and the Operation stays mute until disposed. The test fake did not model the exclusive slot, which is why this stayed green — `createFakeSdkFactory` now rejects a second `openSession()` until `close()` runs.
- P2 `chat-session.ts:1057` — suppressing the second `turn-start` was indeed not enough. `dispatch()` now awaits `awaitTurnClose()` before opening its turn, so a user message that arrives while a background-woken turn is open queues instead of riding that turn's `result`. `dispose()` also releases queued waiters so a session teardown cannot strand `turnFlight`.

## Validation
- `npx vitest run` (terminal plugin) — 893 passed
- `pnpm --filter '@fleet-plugins/terminal' typecheck` — clean
- Both fixes verified load-bearing: reverting `session.close()` fails "closes the retired session so the next message can open a new one"; reverting the await fails "waits for a background-woken turn to close before dispatching the next message"

## Notes
- No declines this pass; both findings reproduced against the current code before editing.

sbluemin:

## Addressed
- P2 `chat-session.ts:433` — confirmed by measurement against the real SDK: a child emits the interrupted turn's `result` about **2ms** after `interrupt()` (`error_during_execution`). `stopTurn()` now marks that pending settlement, `awaitTurnClose()` holds new dispatches behind it, and `ingest()` consumes that one `turn-end` instead of drawing it (if a turn is genuinely open, that turn owns the result instead). `retireSession()` clears the mark so a dead child cannot strand the queue.
- P2 `sdk.ts:183` — `startTurn` and `openSession` now reserve the slot before the first `await` via a placeholder and swap in the real handle, releasing on failure and honouring a `dispose()` that lands mid-preparation. Applied to both entry points since they share one slot.

## Validation
- `pnpm build`, `pnpm typecheck`, `pnpm test` — all clean across the workspace
- Each fix verified load-bearing by reverting it against its own case: slot reservation fails "reserves the slot before awaiting launch preparation" and "keeps a session and a turn out of the same slot even when they race"; the stop marker fails "keeps a new send behind the interrupted turn's result"
- The core-agent SDK mock previously had no `runVendorSession`, so a session/turn race would have passed for the wrong reason; the mock now models it and the race asserts exactly one child is spawned

## Notes
- No declines this pass; both findings reproduced before editing.

sbluemin:
@codex The third-pass feedback has been addressed. Please re-review.

## Addressed
- P2 `chat-session.ts:443` — correct, and it was a defect my previous fix introduced. The marker now tracks whether the prompt actually reached the child: false for a dispatch still waiting on `ensureSession()`, true from the moment `send()` lands, and true from the start for a turn the child opened by itself. A stop before the prompt reaches the child therefore waits for nothing.

## Validation
- `pnpm build`, `pnpm typecheck`, `pnpm test` — clean across the workspace
- Verified load-bearing: reverting the condition hangs "does not wait for a settlement when the stopped turn never reached the child"
- Since three passes in a row found a defect my own previous fix created, I also walked the state axes directly rather than waiting for another pass: dispose/stop while a dispatch is queued, the child dying mid-settlement, a background-woken turn arriving during a settlement, and a send left in the queue after a stop all have a releasing path.

## Notes
- No declines this pass.

sbluemin:
@codex The fourth-pass feedback has been addressed. Please re-review.

## Addressed
- P2 `chat-session.ts:1153` — the dispatch now verifies `this.session === session` before arming its waiter. A child that dies during startup retires the session between `ensureSession()` resolving and this continuation, and the vendor session silently drops sends after close, so the queue had nothing left to release it.
- P2 `chat-session.ts:1184` — requests arriving during a sync flight now mark it dirty and rerun once, guarded on `latestSessionId !== reportedSessionId`. The case that actually bites is a session id changing mid-flight: the swallowed request carried the newer coordinate, and if that frame was the last one the Operation kept the stale one as durable authority. (A late-created transcript is a different matter and stays as before: the next frame or the next turn retries, which is the pre-existing contract.)

## Validation
- `pnpm build`, `pnpm typecheck`, `pnpm test` — clean across the workspace
- Both verified load-bearing against their own cases

## Notes
- No declines this pass.
- Every pass so far has found a real defect, and three of them were created by my own previous fix. I am treating this as the last review pass: if the next one has no FIX-class finding I will run the final context audit and merge, and if it does I will report the state rather than continue the loop indefinitely.

PR #800, opened 2026-08-17
feat(fleet-console): fold status groups into one shelf caption

sbluemin:
## Summary
- Fold every STATUS group in the Map sidebar and War Room into one shelf caption row (name + count + collapse).
- Keep the 3px signal stripe on that row and drop the 8% tinted wash so the column is no longer a colored group block.
- Speak restore/restart verbs only on Minimized and Ended. Living buckets stay names: Awaiting / Running / Idle.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console exec vitest run` on status-axis, group, design-contract, feature-tour, triage-store (287 passed)
- [x] Isolated console: restore moves a chip from Minimized to Idle; re-minimize moves it back; War Room uses `Minimized · select to restore to the deck` with no outer caption stack

### DIFF
--- a/.changelog.d/status-shelf-caption.md
+++ b/.changelog.d/status-shelf-caption.md
@@ -1,0 +1,8 @@
+---
+branch: status-shelf-caption
+---
+
+### fleet-console
+#### Changed
+- Fold every status group in the left sidebar and War Room into one shelf caption that keeps the signal stripe and count, drops the tinted wash, and keeps restore verbs on Minimized and Ended only.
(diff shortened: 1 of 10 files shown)


sbluemin:
Codex review quota is exhausted on this PR (the connector already posted the usage-limit notice). Leaving this note so the skip is visible; I will still request a review once, then continue if activation does not arrive.

sbluemin:
@codex Please review this PR.

PR #850, opened 2026-08-21
feat(ai-gateway): let a panel run its own gateway when the user asks for one

sbluemin:
## Summary

- Settings → AI Gateway gains **패널별 전용 게이트웨이**, an off-by-default toggle placed directly above the *AI Gateway 모델* card. Off keeps today's behaviour exactly: one router per Console, shared by every panel. On, each panel launched afterwards gets its own router — its own upstream in-flight ceiling, its own Cursor adapter and the HTTP/2 sessions it dials, its own withheld-skill memory.
- The panel identity rides the **`x-fleet-panel` request header**, not the URL. Claude Code keeps one gateway model cache per Claude home and accepts it only when the cached `baseUrl` equals `ANTHROPIC_BASE_URL` character-for-character — measured against the installed 2.1.238 binary, a mismatch returns an *empty* gateway model list and performs no live refresh. Panels share a home, so a per-panel base URL would have emptied the `/model` picker for every panel but the most recently launched one. A header changes nothing the cache compares; it is carried via `ANTHROPIC_CUSTOM_HEADERS`, and a CLI too old to support that simply falls back to the shared gateway.
- The pool only ever *looks up* an id it issued, so a caller-supplied header can never mint a router. A panel's router is released when its Operation is deleted, a ceiling of 32 bounds a Console that never releases, and a request naming no live panel is served by the shared router rather than refused — a child that outlived a Console restart keeps working.

## What this does not do

The toggle isolates this process's own bookkeeping and nothing else. Panels still share one credential per provider, one provider-side rate limit, and one failure journal. The help text states this rather than implying extra headroom.

## Test Plan

- [x] `pnpm -r test` — every workspace green (terminal 993, console 2303, core-ai-gateway 855)
- [x] `pnpm build` — clean, including `tsc` for every workspace
- [x] New `tests/ai-gateway-pool.test.ts` (19 cases): claim/resolve/release/dispose, the ceiling, the settings-read failure fallback, header-id validation, and the `ANTHROPIC_CUSTOM_HEADERS` composition (preserves the user's own headers, replaces an inherited panel id)
- [x] Red/green proven by reverting each guard in turn — resolve-mints-on-miss, no ceiling, and re-claim-ignores-existing each fail the intended cases; the launch header injection likewise
- [x] `tests/ai-gateway-routes.test.ts` — the returned pool follows the setting, and host teardown disposes every panel router
- [x] `tests/launch.test.ts` — the header reaches the spawn env while `ANTHROPIC_BASE_URL` and the written cache `baseUrl` stay identical across panels
- [x] `tests/settings-routes.test.ts` — PUT persistence both directions, non-boolean rejection, and survival across an unrelated save
- [x] Headed E2E on an isolated Console: card renders above *AI Gateway 모델*, toggle On writes `dedicatedGatewayPerPanel: true`, survives reload, toggle Off removes the key; no page errors
- [x] `node scripts/compile-changelog-fragments.mjs --check` — 13 entries validated


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed
- **P1 — Route Chat Mode sessions through the panel gateway** (`chat-session.ts:1084-1088`) — confirmed: `ensureSdk()` built its child env from `chatChildEnv(process.env)` alone, so a panel with its own router kept it for PTY turns while its chat turns went to the shared one. The chat seed now carries `panelId`, claimed from the same pool the launch path uses (`agent.ts`), and `chatSdkEnv()` composes the `x-fleet-panel` header into the SDK child environment. An Operation launched as a terminal and later switched to Chat Mode re-claims and stays on the router it already had, because `claim` returns an existing id rather than minting a second router.
- The base URL stays untouched on this surface for the same reason it is untouched on the other: both faces share one Claude home whose gateway model cache compares that string character-for-character.
- Header composition now returns a value (`panelGatewayHeaderValue`) instead of a merged environment. The PTY profile's env and the Chat Mode child's `ProcessEnv` are different shapes, and one merging signature would have had to lie about one of them.

## Validation
- `pnpm -r test` — all workspaces green (terminal 995, console 2303, core-ai-gateway 855)
- `pnpm build` — clean, `tsc` included
- New `agent-chat-session.test.ts` cases: the header reaches the SDK child when the seed carries a panel id, and is absent when it does not
- Red proven: reverting `chatSdkEnv` back to `chatChildEnv(process.env)` fails the first of those two cases
- Cumulative diff from `REVIEW_BASE_HEAD` `7e636d71e` audited against the frozen scope — every hunk serves this one finding or the helper reshape it required; no opportunistic change

## Notes
- No declines or deferrals this pass.

sbluemin:

## Addressed
- **P2 — Preserve the launch-time shared-router decision** (`agent.ts:1378`) — confirmed and fixed at the pool instead of the call site, because the same re-evaluation would reappear at any future third surface. `claim` now records the shared answer as well, so it repeats whatever decision that operation already received rather than re-reading the setting. A ceiling refusal is recorded for the same reason: without it, one freed slot would hand a later surface a dedicated router the first surface is not using.
- A settings read that *threw* is deliberately not recorded. It produced no decision, only a failure to make one, and the next surface deserves a real answer rather than an inherited fallback.
- Remembering shared answers is bounded at `MAX_REMEMBERED_SHARED_PANELS = 512`, since an entry would otherwise outlive every Operation a long-running Console never deletes. Past the bound the oldest is forgotten and that operation is re-decided — which is the very defect this memory closes — so the bound sits far above any believable count of live Operations rather than being tuned close to one. `release` drops the decision with the Operation.

## Validation
- `pnpm -r test` — all workspaces green (terminal 1000, console 2303, core-ai-gateway 855)
- `pnpm build` — clean, `tsc` included
- 5 new pool cases: the off-then-on Chat Mode path, ceiling-refusal stickiness, decision independence between operations, the settings-read-failure exception, release forgetting, and eviction past the bound
- Red proven: dropping the shared record and the two `decideShared` call sites fails exactly the three cases that assert stickiness
- Cumulative diff from `REVIEW_BASE_HEAD` `7e636d71e` re-audited — every hunk still serves the two reported findings; no opportunistic change

## Notes
- No declines or deferrals this pass.

sbluemin:
@codex Reviewed and declined this pass, with one deferral recorded. No code changes.

## Declined
- **P2 — Decouple routing tokens from opaque operation IDs** — I could not reproduce it on any shipped path, so under this repo's judgment policy it stays a hypothesis rather than a fix.
  - Every Operation id on a shipped path comes from `crypto.randomUUID()` (`operations-domain.ts:147,199`), which is always inside the grammar.
  - No production caller passes a custom `id`. The only `input.id ?? ...` producers in the tree are the domain itself and a test harness; the Console client and plugins never send one.
  - Measured the actual durable state on this machine: 4 Operations, all UUIDs, zero outside `[A-Za-z0-9._-]{1,128}`. The premise that restored state carries wider ids did not hold against real data.
  - The failure mode is safe degradation — the panel keeps working on the shared gateway — not corruption, data loss, or a security boundary.
  - Reaching it requires POSTing a hand-made id to the authorized Operations API. That is a state I would have to construct to satisfy the review, not one a workflow produces.

## Deferred
- The coupling itself is real and I want it on the record: the panel token is *derived* from an opaque host id, so my grammar and the host's must agree forever, and a disagreement fails silently — the setting would just stop isolating that panel with no signal anywhere. The right shape is an issued or encoded token, exactly as you describe. It becomes a defect the moment any shipped path issues a non-UUID Operation id; until then it is a maintenance hazard, and absorbing an identity-scheme change into a PR whose scope is a settings toggle would widen it for a hypothetical.

## Validation
- No code changed this pass. The tree is unchanged from `8e33be151`, which is green: `pnpm -r test` (terminal 1000, console 2303, core-ai-gateway 855) and `pnpm build` clean.

This pass produced no FIX-class finding, so I am closing the review loop here and merging.

PR #854, opened 2026-08-21
feat(ai-gateway): run a panel's gateway as its own process

sbluemin:
## Summary

- The dedicated-gateway setting shipped in #850 as another **router object inside the Console**, which was the wrong reading of what a dedicated gateway is. Node keeps **one global undici dispatcher per process** — measured, the first `fetch` installs `Symbol(undici.globalDispatcher.1)` on `globalThis` — so routers sharing a process share every upstream connection pool no matter how many of them there are, and a provider adapter holding module-scoped state shares that too. The isolation the setting advertised stopped at the downstream door.
- A panel now gets a **gateway process**: its own port, its own dispatcher, its own adapter state, its own crash domain. Settings and credentials stay shared, because a panel is meant to get its own connections rather than its own configuration, and the failure journal stays one file so diagnosing a machine remains one lookup.
- The header-routing machinery from #850 is removed. The setting axis and its Settings card are kept, with help text rewritten to describe what it now actually does — including that it costs memory.

## Measured

The child runs its own slim bundle rather than re-entering the Console CLI. On this machine:

| entry | startup | physical footprint |
|---|---|---|
| `dist/panel-gateway.mjs` (1.19 MB) | **95 ms** | **42.5 MiB** |
| `dist/cli.mjs panel-gateway` (2.11 MB) | 222 ms | 84.6 MiB |
| bare Node floor | — | 10.8 MiB |

Loading the server and the plugin host once per panel bought nothing. Where the slim bundle is absent the command falls back to the console entry — a heavier gateway beats no isolation at all. Eight panels is the ceiling, so the worst case is roughly 340 MiB.

## Failure behaviour

Everything that can fail falls back to the shared gateway instead of failing the launch: the setting off, no operation identity, the eight-process ceiling, a child that never reports a port, a child that dies first, a settings read that throws. The answer is decided once per operation and repeated, so a terminal panel later opened in Chat Mode stays on the gateway it already had.

## Known gap

The gateway model cache is one file per Claude home and is compared against `ANTHROPIC_BASE_URL` exactly, so per-port gateways make it single-writer. Measured: Claude Code reads that file **once per child and memoizes it**, so each panel keeps the correct list for its lifetime; the residual race is two panels launched within about a second of each other, where the earlier one can see an empty `/model` picker. Launch serialization closes it and is not in this PR.

## Test Plan

- [x] `pnpm -r test` — every workspace green (terminal 995, console 2307, core-ai-gateway 859)
- [x] `pnpm build` — clean, `tsc` for every workspace
- [x] `tests/panel-gateway-pool.test.ts` (18): ceiling, settings-read failure, child that never reports, child that dies first, concurrent claim from both surfaces, release, teardown, decision stickiness
- [x] `tests/panel-gateway.test.ts` (6): the child listens on its own port, serves `/api/hello`, returns 401 without a credential (proving the router is really mounted), answers nothing outside its route, two children take two ports, and it exits 0 when its parent closes stdin
- [x] `tests/panel-gateway-live.test.ts` (5): command resolution picks the slim bundle and falls back correctly, the pool hands back a URL a real child answers, two panels get two live gateways, and **release actually frees the port**
- [x] Handoff console booted from this branch's build: setting persisted On, Theater registered, no stray gateway processes after shutdown
- [x] `node scripts/compile-changelog-fragments.mjs --check` — 30 entries validated
- [x] `node scripts/check-core-boundary.mjs` — clean

Changelog-Amend: feat-dedicated-gateway-per-panel.md


sbluemin:
@codex Please review this PR.

sbluemin:
@codex Both findings were real and are fixed. Please re-review.

## Addressed
- **P1 — Append the mounted route to the child URL** — confirmed and worse than a missing string: the pool returned `http://127.0.0.1:<port>` while the child mounts at `/ai-gateway`, so a dedicated panel's `ANTHROPIC_BASE_URL` lacked the mount and every model request landed on the child's 404 branch. The feature was broken end to end for its only real use.
  - Fixed structurally rather than by concatenating in the pool: **the child now reports its full base URL** on the ready line. It is the only side that knows where it mounted, and reconstructing that in the parent is exactly what went wrong.
  - The parent validates the reported string is a loopback `http` URL before baking it into a child environment, since a bad value there kills every turn on that panel rather than one request.
- **P2 — Cancel gateways that are released while starting** — confirmed. `release()` found nothing live, `start()` published the child afterwards, and the process outlived its deleted Operation while holding one of the eight slots. A release landing mid-startup is now remembered and the child is shut down when it reports ready, which also returns the slot to the ceiling.

## On the tests that should have caught P1
Three tests covered that path and all three stayed green, because **each appended `/ai-gateway` to the URL itself before fetching** — a test that repairs the value under test cannot fail for the reason it exists. They now dial the URL verbatim the way Claude Code does, one asserts its shape outright (`/^http:\/\/127\.0\.0\.1:\d+\/ai-gateway$/`), and a new case posts to the `/v1/messages` path a client actually appends and expects 401 rather than 404.

## Validation
- `pnpm -r test` — all workspaces green (terminal 1000, console 2307, core-ai-gateway 859)
- `pnpm build` — clean
- **Red proven**: dropping the mount from the child's ready line again fails 4 of the 6 live cases, including the new `/v1/messages` one. Before this commit that same break failed nothing.
- 4 new pool cases: the mounted URL is handed back verbatim, a non-loopback report falls back to the shared gateway, a mid-start release shuts the child down, and the freed slot is reusable
- Cumulative diff from `REVIEW_BASE_HEAD` `643ff1021` audited — every hunk serves these two findings; no opportunistic change

## Notes
- No declines or deferrals this pass.

sbluemin:
@codex One fixed, one declined with evidence. Please re-review.

## Fixed
- **P2 — Make shared log rotation safe across gateway processes** — confirmed. I had pointed panel gateways at the Console's own journal and diagnostic log deliberately, so that "this machine's failures" stayed one lookup, and that choice created the race you describe: both writers serialize rotation only inside one process, so one gateway's `rename` can replace what another just created. Each gateway process now writes its own files beside the Console's; the lookup becomes a directory rather than a file, and the Console's own names are unchanged so an existing habit still finds what it used to. `createCursorDiagnosticLog` gained a file-name option, documented where a caller running several writers over one directory will read it.

## Declined, with the measurement
- **P2 — Serialize launches before rewriting the shared model cache** — the race is real and I documented it in the PR body before you raised it. I am declining the proposed remedies because I measured both and neither closes it.
  - **"Give each panel an isolated cache" is blocked.** The cache path is derived from the config dir alone (`join(configDir, "cache", "gateway-models.json")` in 2.1.238); there is no separate cache-dir variable. Moving `CLAUDE_CONFIG_DIR` per panel makes the child derive a keychain suffix and it dies with `Not logged in` — measured directly while investigating this.
  - **"Serialize the cache-write-and-child-start sequence" does not close it either.** The window that matters runs from our write until *the child* reads the file, and that read happens inside a process we spawned and cannot observe. A lock we hold across our own code would have to be released at spawn, before the read. What remains is a timing heuristic — hold a lock for some milliseconds and hope — and I would rather not add an unbounded sleep to every gateway launch and call the defect closed.
  - **The blast radius is bounded and I measured why.** The reader memoizes per process: `cCd()` caches by path in `gatewayModelsByCachePath` and reads the file once. So each child keeps a correct list for its whole life, and only a child whose read is beaten by the *next* panel's write is affected. The casualty is the `/model` picker's gateway entries for that one panel until it relaunches; the launch model still works because it is passed on argv.
  - It becomes worth fixing if Claude Code ever exposes a cache path independent of the config dir, or if the picker turns out to matter more in practice than the argv model. I would rather ship the honest gap than a sleep that hides it.

## Validation
- `pnpm -r test` — all workspaces green (terminal 1000, console 2308, core-ai-gateway 859)
- `pnpm build` — clean
- New case asserts a gateway child writes no file named `failures.jsonl*`, so a future change that re-shares the journal fails here
- Cumulative diff from `REVIEW_BASE_HEAD` `643ff1021` re-audited — every hunk serves a reported finding

sbluemin:
@codex Both fixed. Please re-review.

## Addressed
- **P1 — Propagate the Console's effective Fleet data root** — confirmed, and it was worse in practice than on paper. The parent now passes its effective root **and the plugin data slot** into the child rather than letting the child rebuild either.
  - I hit this myself without noticing: my measurement runs set `FLEET_CONSOLE_DATA_DIR` *and* `FLEET_DATA_DIR` together, which masked it, while the handoff console I booted for the user set only the former — so a panel launched there would have read the real `~/.fleet` and written diagnostics into it, in the exact run I described as isolated.
  - The log directory is passed rather than derived for the same reason the base URL is now reported rather than reassembled: where a plugin's data lives is the host's answer, and rebuilding another side's path is what dropped the gateway mount two commits ago.
- **P2 — Honor the stored wire-log setting in gateway children** — confirmed, and wrong in both directions as you describe. Turning the log on left every dedicated panel out of it; turning it off did not stop an inherited `FLEET_GATEWAY_WIRE_LOG` from recording prompts and completions the user had explicitly refused. An absent setting still defers to the environment, `false` now overrides it, and a settings read that throws fails closed — the discipline `applyStoredWireLog` already uses. The target is cleared on shutdown.

## On how these are tested
Both decisions are extracted as pure functions and pinned directly. I tried the end-to-end route first and threw it away: **every request cheap enough for a test is refused by the credential or exposure gate before anything writes a file**, so `expect(no file)` passed whether or not the fix was present. Two vacuous tests are worse than none, and this PR has already shipped three of them.

- Red proven on the real assertions: making `panelGatewayLogDir` ignore the host's value, and `panelGatewayWireLogTarget` return `undefined` instead of `null` for `false`, fails exactly those two cases.
- Separately verified by hand that a dedicated gateway process serves a **real turn**: an xAI request through a child on its own port returned 200 with model output, and with `wireLogEnabled: true` it wrote `wire-log.jsonl` into the host-named directory rather than the inherited env target.

## Validation
- `pnpm -r test` — all workspaces green (terminal 1000, console 2313, core-ai-gateway 859)
- `pnpm build` — clean
- Cumulative diff from `REVIEW_BASE_HEAD` `643ff1021` re-audited — every hunk serves a reported finding

## Notes
- No declines this pass. The model-cache race declined last pass stands, on the evidence given there.

sbluemin:
Withdrawn at the user's request. The per-panel dedicated gateway is being dropped entirely, and #850 is reverted on canary along with it.

PR #865, opened 2026-08-22
feat(ai-gateway): strip Anthropic client identity from provider turns

sbluemin:
## Summary

- Claude Code prepends two blocks to `system` that belong to Anthropic alone: an `x-anthropic-billing-header:` telemetry line (client version + entrypoint) and an identity sentence naming the agent as Claude Code or as a Claude Agent SDK agent. Both were forwarded verbatim to every gateway provider, where the telemetry is not theirs to read and the identity is false.
- Adds `stripClientIdentity` to `GatewayPolicySteps`, declared by all six provider policies. It sits in the policies rather than unconditional in `router.ts` because it must **not** reach the Anthropic passthrough, where both blocks are accurate — the mirror image of the usage-limit strip's placement. `packages/core-ai-gateway/CLAUDE.md` now states both halves of that rule.
- Measured against Cloud Code Assist: the identity sentence in its Agent SDK spelling is refused upstream with `429 RESOURCE_EXHAUSTED`, so every Claude Code subagent and every headless `-p` run on Antigravity failed and burned the client's retry budget for ~3 minutes per attempt. That path now works.

Judgment is by block shape, not by exact wording: the billing header carries a version that changes every release, and the identity sentence has two spellings. Both are anchored on the opening they have kept and must be the whole of a single-line block, so a real instruction that merely opens the same way is never deleted.

## Test Plan

- [x] `pnpm --filter @dotobokuri/core-ai-gateway typecheck`
- [x] `pnpm --filter @dotobokuri/core-ai-gateway build`
- [x] `pnpm --filter @dotobokuri/core-ai-gateway test` — 924 passed (41 files); 12 new unit + 8 new policy cases
- [x] Falsification: dropping the step from the xai policy alone fails `strips Anthropic client identity and billing for xai` (1 failed / 923 passed). The case iterates `GATEWAY_PROVIDERS`, so a provider added without the step fails there.
- [x] `node scripts/check-core-boundary.mjs`, `node scripts/check-claude-agent-sdk-boundary.mjs`
- [x] Live upstream A/B with a real captured subagent prompt through the production path (policy -> protocol -> envelope -> Cloud Code Assist): before `429` at 4390 chars of `systemInstruction`, after `200` at 4233 chars.
- [x] End-to-end on an isolated Console built from this branch, real Claude Code against the gateway:
  - Host (interactive TUI, Antigravity pinned): 4 requests, 0 upstream errors, 22,423-token prompt answered.
  - Subagent (headless host dispatching `fleet:antigravity-gemini-3-7-flash-1m-medium`): 6 requests, 0 upstream errors, subagent returned `PONG` and the host reported it.
  - Baseline on the previous build for the same paths: 21/20/9 requests, all `429`.


sbluemin:

## Addressed
- **P2 — Preserve one-line prompts sharing the identity opener** (`packages/core-ai-gateway/src/anthropic/claude-context.ts:1004`) — confirmed, not hypothetical. The old pattern accepted arbitrary text after the opener, so `You are Claude Code. Always answer in Korean and cite file paths.` matched and its whole system block was dropped. The doc comment already claimed the whole-block requirement protected such an instruction; it only did so when the extra text sat on a *following* line, which is exactly the gap you named.

  The matcher now requires the block to be a single sentence that names Anthropic: `[^.\n]*\bAnthropic\b[^.\n]*\.?` — no period before the sentence ends, and `Anthropic` inside it. Both metadata spellings still match and can still be reworded within the sentence; an instruction that merely opens the same way keeps its block, as does the real sentence with instructions appended after it (a block is only ever removed whole).

## Validation
- Reproduced first: the old pattern returned `true` for both one-line instruction cases before any edit.
- Four negative cases added and pinned — same-line CLI opener, same-line SDK opener, identity sentence with text appended, opener continued without naming Anthropic.
- Falsification: reverting only the regex fails exactly those four (`4 failed | 13 passed`); with the fix, `17 passed`.
- `pnpm --filter @dotobokuri/core-ai-gateway typecheck` / `build` clean; `test` 928 passed (41 files), up from 924.
- Live upstream re-check that the real sentence is still stripped: before `429` at 4390 chars of `systemInstruction`, after `200` at 4233 chars -- unchanged by the tightening.

## Notes
- The `x-anthropic-billing-header:` pattern was left as-is deliberately. Its identifying token is the header name itself, so text after it is that header's value rather than a caller's prose; it carries no equivalent over-match. Tightening it too would have been scope the finding did not support.
- Erring toward keeping a block is the intended direction and is now stated in the doc comment: failing to strip a reworded sentence restores a loud symptom (identity leaks, Cloud Code Assist refuses the turn), while stripping a caller's real instruction deletes prompt content silently.

sbluemin:

## Addressed
- **P2 — Preserve single-sentence instructions that mention Anthropic** (`packages/core-ai-gateway/src/anthropic/claude-context.ts`) — confirmed by reproduction: your example matched and its whole block was dropped.

  Taken together with the previous pass, this is the argument against a shape test rather than a request for a third heuristic, so the heuristic is gone. The two identity sentences measured on the wire are now matched **literally**; there is no shape allowance left to widen. Unlike the bracketed usage-limit directive this text is ordinary prose a caller can plausibly write, so any pattern loose enough to survive a rewording is loose enough to swallow a real instruction.

  The `x-anthropic-billing-header:` prefix test stays, and the asymmetry is now explicit in the doc comment: nothing but that header opens with that name, and its client version changes every release, so a prefix test is both sufficient and necessary there. The identity sentence has no such self-identifying token, so it gets none.

## Validation
- Reproduced before editing: the pass-1 pattern returned `true` for your example.
- Three negative cases added and pinned — your counterexample, plus a near-miss of each real sentence (`You are Claude Code, Anthropic's official CLI.` / `You are a Claude agent built on the Claude Agent SDK.`).
- Falsification: restoring the pass-1 shape fails `keeps a near-miss of the CLI sentence`; with the literal set, `20 passed` in that file.
- `pnpm --filter @dotobokuri/core-ai-gateway typecheck` / `build` clean; `test` 931 passed (41 files).
- `node scripts/check-core-boundary.mjs`, `node scripts/check-claude-agent-sdk-boundary.mjs` — both clean.
- Live upstream re-check that the real sentence is still stripped: before `429` at 4390 chars of `systemInstruction`, after `200` at 4233 chars -- unchanged across both fixes.

## Notes
- Going stale is now the **intended** failure mode and is documented as such. An unrecognized rewording restores a loud symptom (identity leaks, Cloud Code Assist refuses the turn) that surfaces on the first run and is fixed by adding the spelling to the list. Deleting a caller's instruction is silent and unrecoverable, so the matcher errs toward keeping a block.
- A length gate precedes the comparison: a `system` block can be an entire skill body, and trimming one to compare it would copy the whole string first.

PR #867, opened 2026-08-22
feat(fleet-console): rebuild Codex as a wiki workbench

sbluemin:
Implements all four directions (A/B/C/D) from the Codex redesign proposal.

## Why

Measured live on `canary eb709a12b`, isolated Console, 1280×577, Instrument, 24-entry Theater:

| | before | after |
|---|---|---|
| Reader: px before the first body sentence | **253 / 358 (70.7%)** | **131 (36.6%)** |
| Reader `h1` in a 530px split pane | **60px** (the *wider* sheet rendered smaller) | **22px**, scales up with the pane |
| Diffs anywhere in the product | **0** | approval gate, conflicts, Cowork |
| Conflict resolution | **impossible in-product** | Keep current / Take proposed / Manual |
| Writing to the wiki | **AI prompt only** | direct Edit + New from template |
| Wiki tools with no route | **7 of 13** | 1 (`compile_source`, see below) |
| Drydock | replayed an agent's log | runs on demand |

The root cause was not styling. fleet-wiki marks its mutation tools *host-only* so no agent carrier can write — correct — and the Console, which **is** the host, never picked them up.

## A — substrate

Codex was a standalone documentation site moved into a 420px rail without re-deriving its layout. Both reader hosts become container-query roots, so document typography follows the pane instead of the window. Codex font tokens become aliases of the console ladder (names kept — shared markdown and core consume them too), off-ladder half-pixel sizes fold onto it, and 98 rule blocks / 845 lines of retired page shell are removed.

## C — review desk

- `DrydockDetailResponse` carries `currentBody`/`currentVersion`; the approval screen renders a unified line diff with collapsed context, `+N -N`, and a `v3 → v4` step. Approve/Reject pinned to the top of the pane.
- Patch sets are grouped in the queue with **Approve all** (`approve_set`, previously unrouted). Partial results report what remains.
- Conflicts are titled with the entry in conflict, render `meta.reason` and `meta.warnings` that were already on the wire and previously dropped, diff current vs proposed, disclose the raw source, and wire `resolveConflict`.
- Direct **Edit** on every entry — a Markdown editor saving through `enqueuePatch` + `approvePatch`, the path Cowork already established. Optimistic concurrency refuses a stale edit rather than overwriting.
- **New entry** from a schema template; **Run check** actually calls `runDryDock()`.

## B — workbench

`/api/entry` returns backlinks (`buildBacklinksIndex` existed, never called) and a link-title map. "Referenced by" is its own section ranked **above** tag similarity, and inline `[[wiki:id]]` links render titles instead of slugs. Measured before: an entry's two authored links ranked 4th and 5th, below three incidental `policy` tag collisions. The navigator gains type/status facets from the enums fleet-wiki already defines.

## D — asking the corpus

A third navigator mode asks through `resolveWikiContext` and answers with the entries carrying the answer, their claims, and the raw sources behind them. It **cites rather than composes**, so no sentence can appear without a source, and `missing_or_uncertain` is shown rather than hidden.

## Doctrine

Cowork painted diff additions with `brass` — the location channel, not a state channel. All three surfaces now share one grammar on `positive`/`coral`. Per doctrine this is a legitimate grammar change, so `instrument-design-contract.test.ts` is updated in the same change (75 tests pass).

## Verification

- 19 new route tests (`codex-review-desk-routes.test.ts`): diff baseline, batch approval, shared write gate, conflict resolution, path containment, stale-edit refusal, duplicate-id refusal, backlinks/link-titles, drydock run, query
- **195 tests green** across the affected suites
- Every flow driven headless end-to-end, including conflict resolve (health count 1 → 0), editor save (v3 → v4, stays on the document), and Run check (2 errors / 26 warnings / 18 info on the real corpus)

## Not included

`wiki_compile_source` (bulk import of one source into a patch set) is the one remaining unrouted tool. It has no exported pure function — only an MCP tool config — so wiring it means adding a public export to fleet-wiki. That is a package-surface change worth its own review, and it is the lowest-value item in the set.

## Correction to the proposal

The proposal claimed "46% of Codex CSS selectors are dead". That was overstated: a static scan cannot see template-composed classes like `` is-${tone} ``. Only 98 blocks were positively confirmed retired and removed. The artifact has been corrected.

Proposal with the full measurement and the to-scale mock: https://claude.ai/code/artifact/a9c01234-4c9f-47ca-b32d-4d318db3fad4


sbluemin:
Closing at the author's request — the Codex redesign is rolled back and not being pursued in this form.

The branch `codex-wiki-proposal` is left on the remote at `6378f388e` so nothing is lost; `canary` was never touched by this work.

PR #878, opened 2026-08-23
feat(core-ai-gateway): add a second harness profile and its path selector

sbluemin:
## Summary

Grok Build launch support is **on hold**; this PR keeps only the `core-ai-gateway` half, which stands on its own. No other package is touched.

- **A second harness is serveable.** `grok-build` profile with its own dotless model id grammar (`grok-build-gateway--`), guarded by a test that the dot-to-dash mapping over the whole catalog stays injective. Selected by a declared path segment (`/ai-gateway/grok/v1/messages`), never by a judgment about the caller — `deps.harnesses` keys are selector segments and an unknown segment falls to the default.
- **`resolveSessionIdentity` on the harness contract.** A client that carries its conversation identity in a header now reaches the `metadata.user_id` both Cursor and Codex derive their session from. A body value always wins, so Claude Code behavior is byte-identical.
- **The Anthropic discovery envelope moved to the wire.** Both harnesses render into it through an injected id renderer, so neither owns the other's dialect and the wire stays free of both.
- **Three Antigravity request-shape fixes**, each measured against a real second client and each a pre-existing defect Claude Code never triggered.

## The Antigravity fixes

A client sending shapes Claude Code does not send made the whole turn die with a 400 before a token was produced:

| Shape | What happened |
|---|---|
| `type: ["string", "null"]` | `Proto field is not repeating, cannot start list`. The key whitelist could not catch it — the key is legal, the value shape is not |
| `{ type: "object" }` with no `properties` | Free-form payload. Gemini refuses it rather than reading it as "any object". Generic 400, names nothing |
| `max_tokens: 128000` | Context window and output ceiling are different numbers: the catalog window here is 1M, the wire refuses output above 64k |

Measured 2026-08-23 against `daily-cloudcode-pa.googleapis.com`: 65,536 accepted, 131,072 refused, with and without tools. Output is now clamped rather than refused — a caller cannot receive more than the provider will emit.

## Scope note

The `grok-build` profile ships **unreferenced by any host**: nothing in this repo passes `deps.harnesses` yet. It is groundwork plus the live proof that the profile seam serves a client answering every field differently. Say the word and I will drop it, leaving the selector, the identity field, the envelope move, and the Antigravity fixes.

## Test Plan

- [x] `core-ai-gateway` 981 tests (+19)
- [x] Consumers unchanged: `fleet-admiral` 194, `terminal` 959, `core-agent` 133
- [x] Repo-wide `pnpm -r typecheck` clean; `check-core-boundary` passes
- [x] `compile-changelog-fragments --check` passes; no fragment, since nothing here is a user-visible delta on its own
- [x] Antigravity fixes verified end to end against the live provider before the rollback: a client sending all three shapes went from a 400 to a completed turn


sbluemin:
@codex The PR scope was reduced. Please re-review the current head.

The branch was force-pushed and now contains **only `packages/core-ai-gateway`**. Every file your previous review commented on (`packages/fleet-admiral/src/agent-cli/grok/*`) has been rolled back and is no longer part of this PR — Grok Build launch support is on hold.

## What remains
- A second harness profile (`grok-build`) with its own dotless model id grammar, plus a path-segment harness selector on the router. No host wires it yet; it is groundwork and the live proof that the profile seam serves a client answering every field differently.
- `resolveSessionIdentity` on the harness contract, so a client carrying its conversation identity in a header reaches the `metadata.user_id` Cursor and Codex derive their session from. A body value always wins.
- The Anthropic discovery envelope lifted from the Claude harness folder into the wire, rendered through an injected id renderer.
- Three Antigravity request-shape fixes: `type` unions collapsed to Gemini's scalar + `nullable`, a declared object with no property map given an empty map, and `maxOutputTokens` clamped to the 65,536 this wire accepts.

## Validation
- `core-ai-gateway` 981 tests, repo-wide `pnpm -r typecheck` clean, `check-core-boundary` passes
- Consumers unchanged: `fleet-admiral` 194, `terminal` 959, `core-agent` 133

sbluemin:

## Addressed
- **P1 — Check own properties when resolving harness selectors** (`packages/core-ai-gateway/src/router/router.ts:236`): FIXED. Reproduced first — `/ai-gateway/toString/v1/messages` resolved to `Object.prototype.toString` and the next line threw `TypeError: harness.acceptsCredential is not a function`. The selector map is now a real `Map`, so an unregistered segment falls to the default harness as documented. A regression test pins that path returning 401 from the default harness instead of crashing.

## Validation
- `pnpm vitest run` (core-ai-gateway) — 982 passed (+1)
- `pnpm typecheck` (core-ai-gateway) — clean
- `pnpm -r typecheck` — clean, 0 errors
- `node scripts/check-core-boundary.mjs` — passes

## Notes
- Cumulative diff from the review base is 2 files, +24/-3, entirely this one finding.

PR #917, opened 2026-08-26
feat(fleet-console): make Codex full-size reading a place to read

sbluemin:
## Summary
- 격리 콘솔 실측으로 확대 읽기의 기능 결함 9건을 고정한 뒤, 사용자 재가를 받은 A~E 전 범위를 인도합니다.
- **자리 보존**: split↔확대 relocate 때 스크롤 스파이를 새 스크롤 루트에 다시 걸고(목차가 첫 섹션에 얼어붙던 결함), 읽기 기록과 새로고침이 읽던 자리를 지키며, 페이지 이탈 시 디바운스된 세션 기록을 flush 합니다.
- **키보드 정독**: 열리는 순간 본문이 초점을 받아 Space·PageDown·Home·End가 먹고, J/K 섹션 점프, ⌘[·⌘] 기록 이동, ⌘F 문서 내 찾기(CSS Custom Highlight)를 더했습니다.
- **문서 사이 이동**: 헤드바가 화면 이름 대신 문서 제목·진행률을 말하고, ⌘K 항목 전환기가 레일과 같은 카탈로그를 읽으며, 본문 태그는 화면 밖 레일을 조용히 거르는 대신 그 태그로 걸러진 전환기를 엽니다.
- **주소화**: 읽는 문서와 확대 여부가 주소에 실려 새로고침·브라우저 뒤로/앞으로·링크 복사가 같은 문서와 화면으로 수렴합니다.
- **정독 도구**: 읽기 폭·크기 3단, 마크다운 원문 보기·복사, 넓은 화면에서 관련 항목·백링크를 본문 옆으로 옮겼습니다.
- 시트가 열려 있는 동안 전역 ⌘K(세션 검색)는 양보해 두 표면이 함께 열리지 않습니다.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console build` — green
- [x] `tsc --noEmit -p core/client/tsconfig.json` — green
- [x] `vitest run tests/instrument-design-contract.test.ts tests/i18n.test.ts tests/codex-panel-state.test.tsx tests/codex-cowork-ui.test.ts tests/codex-schema-ui.test.ts` — 135 passed
- [x] 격리 콘솔 실측(1920×977, 헤드리스 실브라우저): 확대 직후 목차 전이 재개 · PageDown/End/Home 본문 이동 · 링크 이동 후 뒤로 3600px 복원 · 새로고침 후 확대+2400px 복원 · 딥링크 진입 · 브라우저 뒤로/앞으로 문서 왕복 · ⌘K 전환기 단독 개방 · 태그 칩→필터된 전환기 · 원문 5,752자 · 넓은 화면 백링크 옆단 배치 · 페이지 오류 0


sbluemin:

## Addressed
- P1 "Wait for the resolved workspace before mounting the reader" — FIX. `rail/codex-panel.tsx`: the resolved workspace is now published to subscribers (`subscribeCodexWorkspace`); `components/codex-reading-sheet.tsx:57,137,186`: the sheet reads it through `useSyncExternalStore`, returns early while it is null instead of falling back to the Theater id, and depends on it so resolution re-runs the mount. Reproduced first: on a cold deep link the entry request left at 175ms carrying the Theater id, before resolution. After the fix it fires once at 250ms with the resolved workspace id.
- P1 "Close the reader when browser history removes its URL" — FIX. `hooks/use-codex-reader-url.ts:64`: dropping the `codex` parameter now calls `closeCodexReader()` instead of `collapseCodexReader()`. Reproduced first: with only the collapse, the reader kept the entry and the reader-to-address effect pushed it back into the URL. After the fix, one Back from the expanded reader leaves `location.search` empty with no sheet, no split reader and no reader host, and the URL stays empty.

## Validation
- `pnpm --filter @dotobokuri/fleet-console build` — green
- `tsc --noEmit -p core/client/tsconfig.json` — green
- `vitest run tests/codex-panel-state.test.tsx tests/codex-cowork-ui.test.ts tests/codex-schema-ui.test.ts tests/i18n.test.ts tests/instrument-design-contract.test.ts` — 135 passed
- Isolated console (1920x977, real browser): deep link still opens full size with one entry request; Back leaves the reader entirely; scroll spy still transitions after expand (4 transitions); body keeps initial focus.

## Notes
- No declines or deferrals in this pass.

sbluemin:

## Addressed
- P1 "Encode the theater in copied reader links" — FIX. `hooks/use-codex-reader-url.ts`: the reader URL now carries `codexTheater`, and `components/codex-reading-sheet.tsx` Copy link writes it. Opening a link for another Theater switches to it when this console knows that Theater; when it does not, the reader stays closed and the pending target is dropped, so the link never resolves the id against the wrong workspace and the attempt does not repeat every render. Verified: a Theater-qualified deep link opens with one entry request; an unknown `codexTheater` leaves the reader closed with the active Theater untouched and no page errors.
- P2 "Update the existing reader-controller test mock" — FIX, and thank you: `tests/codex-host-state.test.ts` was failing 11 of 13 and I had not run that file. The mock now defines `refreshScrollSpy` and `getDocument`, and the persisted-session expectation records `expanded: false`, which is the field this PR adds to the reader session.

## Validation
- `pnpm --filter @dotobokuri/fleet-console build` — green
- `tsc --noEmit -p core/client/tsconfig.json` — green
- `vitest run tests/codex-host-state.test.ts tests/codex-panel-state.test.tsx tests/codex-cowork-ui.test.ts tests/codex-schema-ui.test.ts tests/codex-queue-actions.test.ts tests/i18n.test.ts tests/instrument-design-contract.test.ts` — 163 passed
- Isolated console (real browser): Theater-qualified deep link opens full size, scroll spy transitions 7 times after expand, entry switcher opens with the same 32-row catalog, body keeps initial focus.

## Notes
- No declines or deferrals in this pass.

sbluemin:
@codex The third-pass feedback has been addressed. Please re-review.

## Addressed
- P1 "Restore saved scroll for URL-targeted entries" — FIX, and it was worse than reported: the position was also being saved as 0. Reproduced: with the document at 2400px, the stored session read `scrollTop: 0`, because the page-hide flush sampled the scroll container while the page was already tearing down. `codex-host.ts` now remembers the last position the scroll listener observed and the flush falls back to it. For restoration, `prepareReaderSessionScroll(theaterId, entryId)` is called on the URL path before opening the document, so the saved position is consumed when the stored session names that same entry — the rail-panel path only runs while no reader exists, which the address beats. Verified: scroll to 2400 -> session records 2400 -> reload -> the reader returns at 2400.
- P2 "Publish the cleared document state" — FIX. `reading-controller.ts` gains `onDocumentChanged`, fired when the current document is cleared, and `codex-host.ts` wires it to `emitDocumentState`. Verified: switching to drydock inside full size drops the head bar back to "Codex · Reading" with Copy link and Source disabled, instead of naming the previous entry.
- Follow-on from the second pass, found while reproducing the above: a link naming an unknown Theater left the previous document on screen under an address naming another Theater. Dropping the target now also closes the reader, so the address and the screen agree.

## Declined
- P1 "Suppress the stale URL write after closing the reader" — DECLINE, not reproduced. I instrumented `history.pushState`/`replaceState` and walked the exact path: open a document, follow a related card (1 push), then Back twice. The log shows a single push (the document move) and no write during either Back; `location.search` ends empty with no sheet, no split reader and no reader host. Structurally, the reverse effect's dependencies are `[reader, expanded, theaterId]` — a Back that only changes `location.search` does not re-run it, so the pre-close `reader` closure is never reached; by the time the close re-renders, `reader` is null and the write is a no-op. If you can name a sequence where the alternating entries appear, I will re-open it.

## Validation
- `pnpm --filter @dotobokuri/fleet-console build` — green
- `tsc --noEmit -p core/client/tsconfig.json` — green
- `vitest run` over the 7 Codex/i18n/design-contract suites — 163 passed
- Isolated console (real browser): reload restores 2400px; drydock switch clears the head bar; unknown-Theater link closes the reader and the address self-corrects; scroll spy still transitions after expand (4); body keeps initial focus.

sbluemin:
@codex The fourth-pass feedback has been addressed. Please re-review.

## Addressed
- P1 "Reuse saved history when applying browser Back" — FIX. Reproduced exactly as described: read A to 1500px, follow a related card to B, browser Back — A came back at 0 and the sheet's arrows still read `[back on, forward off]`, i.e. a duplicate `{A, 0}` had been appended. `codex-host.ts` gains `stepReaderHistoryTo(entryId)`, and the URL hook takes that step when the target sits immediately behind or ahead of the current index, falling back to a fresh open otherwise. Verified: Back returns A at 1500px with arrows `[off, on]`; Forward returns B with `[on, off]`.
- P2 "Seed history with the session-restored entry" — FIX. `mountReaderInto` now seeds the restored document as the first history entry with its restored position (only when the stack is empty, so it never competes with a real visit). Verified: after a restored session, moving to another entry leaves Back enabled, where before both arrows stayed disabled.

## Validation
- `pnpm --filter @dotobokuri/fleet-console build` — green
- `tsc --noEmit -p core/client/tsconfig.json` — green
- `vitest run` over the 7 Codex/i18n/design-contract suites — 163 passed
- Isolated console (real browser): Back/Forward carry position and direction; restored session seeds history; scroll spy still transitions after expand (4); body keeps initial focus; no page errors.

## Notes
- This is the fourth review pass. Both findings were reproduced first, so the pass was worth taking; unless the next pass reports a defect I can reproduce, I intend to close the loop here and merge after the final context audit.

sbluemin:
@codex The two P2 findings are fixed; the P1 is declined for the third time, with the same evidence.

## Addressed
- P2 "Distinguish readers with matching IDs across theaters" — FIX. The target now keeps its Theater provenance through the match: `hooks/use-codex-reader-url.ts` returns instead of clearing the target while `target.theaterId` differs from the active Theater, so a link for another Theater's same-id entry survives the switch instead of being erased by the reverse sync.
- P2 "Preserve source offsets while folding search text" — FIX. `collectRanges` now folds with `foldKeepingLength`, which leaves a character as-is when its lowercase form has a different UTF-16 length, so folded indexes always address the original node. Verified: a text node reading `İfoo bar` matches `foo` 1/1 with no page errors; before, `setEnd` would have thrown.

## Declined (third pass on the same claim)
- P1 "Suppress the reverse write during URL-driven close" — DECLINE. The claim is that the reverse effect runs in the same passive-effect flush with the pre-close `reader`. It does not: its dependency array is `[reader, expanded, theaterId]`, and a Back that only changes `location.search` leaves all three identical, so React skips the effect entirely — there is no same-render closure to suppress. By the time the close commits, `reader` is null and the write is a no-op against an address that already has no `codex`. Measured twice with `history.pushState`/`replaceState` instrumented across the exact sequence (open A, follow a card to B, Back, Back): one push total, the document move, and zero writes during either Back; the address ends empty with no sheet, no split reader and no reader host. I am closing this thread unless you can name a concrete sequence that produces the alternating entries.

## Validation
- `pnpm --filter @dotobokuri/fleet-console build` — green
- `tsc --noEmit -p core/client/tsconfig.json` — green
- `vitest run` over the 7 Codex/i18n/design-contract suites — 163 passed
- Isolated console (real browser): Back returns the previous document at 1500px with arrows flipped, scroll spy transitions 6 times after expand, deep link intact, no page errors.

This is the fifth review pass. Both reproducible findings are fixed, so I am closing the loop here and merging after the final context audit.

PR #957, opened 2026-08-30
feat(fleet-console): host-owned pane surface contract for the rail

sbluemin:
## Summary

- 우측 레일과 확대 화면의 레이아웃을 호스트 소유 계약으로 승격했다. **표면은 그릇, 페인은 열**이며, 기하·분할선·캡션·포커스는 호스트가 소유하고 플러그인은 캡션 아래 본문만 그린다. 레일과 확대 캔버스는 같은 어휘의 두 마운트다.
- 계약이 감사에서 드러난 선결 조건을 스스로 닫는다: `keepAlive`로 닫기가 언마운트를 뜻하지 않게 하고, 인스턴스별 `AbortSignal`로 닫힌 페인의 응답이 후속 페인에 착지하는 것을 막고, 팔레트 결과가 모듈 싱글턴에 타깃을 적는 대신 `PaneTarget`을 반환한다.
- **확대가 공통 동작이 됐다.** `mounts`에 `"expanded"`를 적은 detail 페인은 캡션에 확대 버튼을 얻고, 호스트 내장 표면이 같은 `PaneBody`를 캔버스 위에 세운다. 페인은 확대를 위한 코드를 쓰지 않고 `ctx.mount`로 자기 위치만 안다.
- 캡션은 role이 정한다 — **detail에만** 선다. primary는 표면과 수명을 같이해 닫거나 확대할 대상이 아니고, 확대 표면에서는 호스트 슬롯 머리가 이미 제목과 닫기를 그린다.
- 내장 플러그인 7종 전부를 `railEntries` + `panes`로 옮겼다. `railPanels` 등록은 하나도 남지 않았고, 호환 투영은 외부 플러그인용으로만 남는다.
- `.history-detail-pane`을 `.history-inspector-shelf`로 개명했다. 가로 분할선 아래 붙는 행이 스스로를 열이라 부르고 있었고, `Pane`이 이제 예약어다.

제품 동작은 이관 전과 같다. detail 페인이 아직 하나도 등록되지 않았기 때문이며, 2단을 자체 구현한 셋(codex · file-explorer · repository)의 열 분해가 후속 작업으로 남는다.

## Test Plan

- [x] `tsc --noEmit` (fleet-console) — clean
- [x] Console 스위트 2326 passed (`instrument-design-contract` 93 포함)
- [x] codex 176 · repository 441 · file-explorer 241 · skills 125 · ledger 125 · quota 52 · terminal 1127
- [x] `pnpm --filter @dotobokuri/fleet-console build` — green
- [x] 격리 콘솔 핸드오프로 전 레일 패널 육안 확인 (캡션이 primary에 서지 않는 것)
- [ ] detail 페인의 캡션과 확대 버튼 — 등록된 detail 페인이 없어 후속 PR에서 검증


sbluemin:

## Addressed
- **P1 — palette lost every migrated panel** (`app.tsx:131`) — the palette now reads the same binding registry the rail does. Entry supplies id/title, its panes supply search, and a `PaletteSearchPanel` shape lets the old and new contracts sit in one list. Reproduced first: `registry.railPanels` is empty after the migration, so `buildPaletteCommands` and `searchRailPanels` both received nothing.
- **P2 — `close(otherPaneId)` opened instead of closing** (`pane-body.tsx:105`) — the fall-through called `onOpen` with empty params, so the pane a caller asked to close was made visible and its target discarded. It now routes to the store's close.
- **P2 — pane id namespace** (`plugin-registry.ts:171`) — corrected the contract rather than the code. Expanded surfaces already rejected prefixing for a stated reason (a plugin calling `open`/`close` with its own local id would not match the promoted id), and panes deliberately follow that rule. The doc now says what the host does: console-wide unique, first registration wins, duplicates warned and dropped.

## Validation
- `tsc --noEmit` (fleet-console) — clean
- Console suite — 2326 passed
- codex 176 · file-explorer 241 · repository 441 · skills 125 · ledger 125 · quota 52 · terminal 1127
- `pnpm --filter @dotobokuri/fleet-console build` — green

## Notes
- Two defects surfaced while fixing the above and are included: reading the registry now tolerates a shape without the new fields (a composition that has not caught up loses that contribution instead of the whole rail), and `pane-body`'s `onContext` callback fed a state update that rebuilt the context it reported — right-rail's suite went from under a second to a 244s hung worker. The context is now built by a hook that caption and body share.
- The `PaneTarget` returned by pane search is still unconsumed by the palette. That landing path is a separate change; taking it halfway would put two landings on one request while the plugin-side singleton still records its own.

sbluemin:

## Addressed
- **P1 — signal died under StrictMode** (`pane-body.tsx:81`) — confirmed `main.tsx:97` wraps the app in `StrictMode`. The ref-held controller carried the rehearsal's abort into the second setup, so a body passing `ctx.signal` to `fetch` failed on its first request in development. The controller is state now: a dead one is replaced and the context rebuilt.
- **P1 — panes outlived their theater** (`rail-surface.tsx:185`) — instances are keyed by `theaterId`, so a theater switch tears the pane down instead of handing it a new id while its controller stays live. This is the hazard the contract claims to prevent, so the implementation had to move rather than the doc.
- **P2 — expanded opens were dropped** (`pane-store.ts:61`) — `panes.open` now resolves the target against the pane index; an omitted mount takes the descriptor's first, and expanded requests go through the expanded surface — the same door the caption's expand button uses.
- **P2 — sibling close ignored keepAlive** (`rail-surface.tsx:18`) — the sibling path now reads the target descriptor and parks it exactly as `handleClose` does for the pane itself.
- **P2 — entry switch left stale detail panes** (`right-rail.tsx:136`) — `resetSurfacePanes` had no call site. The surface clears on entry change, keeping only declared keep-alive panes.

## Validation
- `tsc --noEmit` (fleet-console) — clean
- Console suite — 2326 passed
- codex 176 · file-explorer 241 · repository 441 · skills 125 · ledger 125 · quota 52 · terminal 1127
- `pnpm --filter @dotobokuri/fleet-console build` — green

## Notes
- No declines this pass; every finding reproduced against the code before it was fixed.

sbluemin:
@codex Addressed. This is the third review pass, so per the workflow's stop rule I am proceeding to the final context audit and merge after posting these dispositions.

## Addressed
- **P1 — signal still dead under StrictMode** (`pane-body.tsx:83`) — you were right that the state replacement lands too late: the rehearsal's cleanup-setup pair happens between renders, so the body's own `[]` effect restarts before the new controller arrives. The cleanup now installs the next controller in place and the context reads it through a getter.
- **P2 — palette discarded the returned target** (`operation-search.tsx:210`) — reversed my earlier decline. A provider that follows the contract and performs no singleton side effect simply never opened. The target is consumed only when present, so the legacy path is untouched.
- **P2 — standalone panes missing from the index** (`pane-registry.ts:172`) — the index is built from the registry now, not from rail bindings, so a pane registered without a rail entry resolves.
- **P2 — retention used the entry being entered** (`rail-surface.tsx:64`) — retention follows each pane's own descriptor, so a keep-alive detail owned by the entry being left survives.
- **P2 — expanded mount's `panes.open` was a no-op** (`expanded-pane-surface.tsx:107`) — routed through the pane index like the rail host.

## Declined
- **P2 — `ctx.width` is not a measured width** (`rail-surface.tsx:208`) — declined as a code change, fixed as a contract statement. Carrying the rendered width would rebuild the context every frame and redraw the body; that is exactly what made `right-rail`'s suite go from under a second to a 244s hung worker two passes ago. Width-reactive layout is the container query's job — the host sets `container-type: inline-size` on the body — so the doc now says what the number is instead of promising a measurement it cannot keep.

## Validation
- `tsc --noEmit` (fleet-console) — clean
- Console suite — 2326 passed
- codex 176 · file-explorer 241 · repository 441 · skills 125 · ledger 125 · quota 52 · terminal 1127
- `pnpm --filter @dotobokuri/fleet-console build` — green

PR #973, opened 2026-08-31
feat(fleet-console): give Session Analyst artifacts a design floor

sbluemin:
## Summary

- 분석가 시스템 프롬프트는 "리터럴 고대비 색을 박고 Console 테마 변수에 의존하지 말라"고 지시하고, `publish_artifact` 도구 description은 "그 변수를 폴백과 함께 선호하라"고 지시해 서로 모순이었습니다. 프롬프트가 이겨 모든 아티팩트가 자기 회색 팔레트를 발명했고, Console 안에서 이질적인 페이지로 읽혔습니다.
- 호스트가 Console 테마를 `--fleet-*` 토큰(면·상태·포커스·타입)으로 넘기고 제목·목록·표·코드·details·근거 인용 칩의 조판 바닥을 깝니다. 모델이 CSS를 한 줄도 쓰지 않아도 아티팩트가 읽힙니다.
- 프롬프트의 리터럴 색 지시를 `# Artifact design` 계약으로 교체했습니다: 토큰+리터럴 폴백만 사용, `prefers-color-scheme` 분기 금지(Console 테마는 OS 테마와 독립), 근거 id는 `<cite>` 칩, 훑어보는 정보 설계, 네트워크 요청 금지.
- 포커스 링을 signal 토큰이 아니라 brass 위치 채널에서 가져와 Console 색 채널 문법을 지킵니다.
- `Download HTML`이 테마와 스타일을 지닌 자립 문서를 저장합니다. `Copy source`는 그대로 분석가 원본을 줍니다.

## Test Plan

- [x] `pnpm --filter @dotobokuri/fleet-analyst test` (36 passed) / `typecheck`
- [x] `pnpm --filter @fleet-plugins/terminal test` (1133 passed) / `typecheck`
- [x] 격리 Console에서 실제 세션분석가(`/brief`, AI Gateway Sonnet)로 아티팩트 발행 후 실측: 주입 토큰 12종 확인, 본문 맨 hex 0개(전부 `var(--fleet-*, #fallback)`), `<cite>` 26개·맨 `[e#]` 나열 0건, `prefers-color-scheme` 0건, 원격 출처 0건
- [x] 다크(instrument)·라이트(whites) 두 테마 렌더 스크린샷 대조, 패널 실폭 422px에서 표 가로 넘침 없음(`pageOverflowX: false`), 페이지 에러/거부 0건


sbluemin:

## Addressed
- P2 Enforce the offline artifact contract in the sandbox — `analysis-routes.ts:16` — FIX. The guarantee was declared in this PR's prompt but left advisory. `ANALYSIS_ARTIFACT_CSP` now closes every remote scheme (`https:`/`http:` removed from `default-src`, `script-src`, `style-src`, `img-src`, `font-src`, `frame-src`, `media-src`) and sets `connect-src 'none'`. `'self'`, `data:`, and `blob:` stay open, so inline CSS, inline scripts, inline SVG, and `data:` images keep working; `frame-ancestors 'self'` stays so the Console can still frame it.
- Downloaded copy losing the header — `analysis-routes.ts:44` — FIX. The document now carries the same resource policy in a `<meta http-equiv="Content-Security-Policy">`, placed before the base stylesheet. `sandbox` and `frame-ancestors` are filtered out because `meta` does not support them.
- The prompt now states the restriction is enforced rather than advised, so the model reaches for an inline asset instead of a remote one.

## Validation
- `pnpm --filter @fleet-plugins/terminal test` — 1133 passed
- `pnpm --filter @dotobokuri/fleet-analyst test` — 36 passed
- `typecheck` for `@fleet-plugins/terminal`, `@dotobokuri/fleet-analyst`, `@dotobokuri/fleet-console` — all clean
- `pnpm --filter @dotobokuri/fleet-console build` — clean
- Live re-check on the hardened build: an isolated Console published a real analyst artifact and it still renders (inline styles intact, 17 citation chips, 0 remote references in the body). Response header carries the closed policy, the document carries the meta policy, and the browser reported 0 page errors and 0 CSP violation logs.

## Notes
- Remote assets were previously allowed and are now blocked. That narrowing is the intended trade-off of this PR, already approved by the owner: an artifact is process-memory only and must not signal outward that a session is under analysis, so it ships inline assets and a system font stack rather than a CDN face.

sbluemin:

## Addressed
- P2 Block form submissions in downloaded artifacts — `analysis-routes.ts:44` — FIX. Correct: `form-action` does not fall back to `default-src`, and the downloaded copy has no response-header sandbox to stop a navigation. `form-action 'none'` now sits in the shared directive list, so it covers both the served header and the document meta policy.
- P2 Do not export raw HTML when wrapping fails — `analysis-artifacts-panel.tsx:115` — FIX. The fallback was introduced by this PR and did recreate the leak: the author's raw HTML carries neither the meta policy nor the base stylesheet. Download now declines instead of falling back, and reports it in place as `Download unavailable` on the menu item rather than failing silently. `Copy source` remains the path to the original HTML.

## Validation
- `pnpm --filter @fleet-plugins/terminal test` — 1134 passed (adds: export refuses the raw fallback and surfaces the notice; the notice clears on menu reopen; header and meta both carry `form-action 'none'`)
- `pnpm --filter @dotobokuri/fleet-analyst test` — 36 passed
- `typecheck` for `@fleet-plugins/terminal` and `@dotobokuri/fleet-console` — clean; `pnpm --filter @dotobokuri/fleet-console build` — clean
- `node scripts/check-localized-attributes.mjs` — no violations (new `terminal.artifacts.exportFailed` key added in both locales)
- No live browser re-run this pass: the change adds one CSP directive and alters only the download failure path, neither of which touches the render surface that was measured on the previous pass.

## Notes
- Artifact forms are not a supported capability — an artifact is a static analysis document — so denying `form-action` narrows nothing a user relies on.

sbluemin:
@codex Third pass reviewed. One finding, declined with rationale; the enforcement boundary is now documented in code.

## Declined
- P2 Block navigation-based outbound requests — `analysis-routes.ts:46` — DECLINE.
  - The underlying facts are correct: CSP has no top-level navigation directive (`navigate-to` was removed from the spec), and a `meta` policy cannot carry `sandbox`. So `<meta http-equiv="refresh">` and a `location` assignment do remain possible in a downloaded copy.
  - The proposed remedy does not pass the scope gate. Navigation cannot be denied while inline script still runs — `location`, `window.open`, `history`, and an anchor click are all reachable — so the only effective measure is stripping scripts from the exported document. Inline `<script>` and `<canvas>` are supported artifact capabilities that predate this PR: the system prompt offers them and the served `script-src` keeps `'unsafe-inline' 'unsafe-eval'`. Removing them on export would turn any artifact carrying a chart or an interactive diagram into a dead file, which narrows shipped behavior for a hypothetical this PR did not introduce.
  - Not reproduced. The path needs an injected transcript, a model that follows it, a user who downloads that artifact, and that user opening the saved file. The last step is a person opening local HTML they exported themselves, which is outside the Console's isolation boundary.
  - What this PR does own is the claim. The meta policy was described as preserving the offline guarantee wherever the file is opened, and that overstated it. `e5c0df4aa` records the real boundary beside the policy: it denies resource loads and form submissions, it cannot deny navigation, the surface needing full isolation is the preview where the response-header sandbox denies everything, and the downloaded file is defense in depth.

## Validation
- `pnpm --filter @fleet-plugins/terminal test` — 1134 passed; `typecheck` clean. Comment-only change, no behavior delta.

## Notes
- Three review passes are complete. Passes 1 and 2 produced FIX-class findings and were applied in full; this pass produced none that survive the scope and reproduction gates, so the review loop closes here per the workflow stop rule.

PR #1010, opened 2026-09-06
feat(console): add opt-in experiments with per-feature model choice

sbluemin:
## Summary
- Add an **Experiments** settings category with four opt-in features that all start off: prompt refinement (a button beside the Quick Launch prompt rewrites the request into a task brief draft the user applies, discards, or restores), a launch context pack (Wiki entries and recent commits offered before launch, appended only when checked), session watch (an Analyst review after each finished turn with results shown in a bubble under the Watch button), and Console reading for Scuttlebutt aides (read-only Operation/Theater/Wiki tools that re-query on every turn).
- Prompt refinement and session watch carry their own model picker (Claude aliases by default, AI Gateway models added by the Terminal plugin); Console reading follows the aide's default model. Both experiments get a first-appearance feature tour.
- The Experiments page also hosts the Quaker aides card (Console reading as one row, roster as multi-select pills) and the Console port / remote access cards; the Experiments chip explains itself on hover, and the AI extensions card keeps each feature on a single line.

## Test Plan
- [x] `pnpm run typecheck` in fleet-console, terminal, scuttlebutt (0 errors)
- [x] fleet-console 332 / terminal 154 / scuttlebutt 8 vitest cases pass
- [x] fleet-console, terminal, scuttlebutt builds; `check-localized-attributes` clean
- [x] Isolated Console (console-e2e / console-handoff): settings page, refinement route, context pack card, aide Console tools, session watch pipeline and bubble exercised in a real browser


sbluemin:

## Addressed
- P1 Chat-mode session watch — `terminal/server/agent-api/chat-session.ts` `closeTurn` / `agent.ts` seed — ChatSession now calls a new `onTurnEnded` seed hook when a turn that reached the child closes (not on user stop), and the Agent route wires it to the same `deps.onTurnEnded`, so watch reviews run in Chat Mode too.
- P1 Console tools after opt-out — `scuttlebutt/server/console-tools.ts` — every Console tool re-reads `ctx.host.experiments.read().aideConsoleRead` at call time and returns `console_read_disabled` once it is off, so a session started while enabled stops reading Console data immediately.
- P2 Launch-context timeout — `core/client/src/launch-experiments.ts` — each provider is raced against the abort signal, so a provider that ignores the signal settles as rejected at the timeout and Quick Launch never stays in `contextPending`.
- P2 Overlapping experiment saves — `sdk/plugin/types.ts` (`experiments.saving()`), `core/client/src/plugin-capabilities.ts`, `scuttlebutt/client/settings-section.tsx` — the Console-reading row is disabled while the core is saving the `experiments` field, matching the core rows, instead of dropping the toggle.

## Validation
- fleet-console `pnpm run typecheck` + vitest 332 passed; `pnpm build` OK
- terminal `pnpm run typecheck` + vitest 154 passed; `pnpm build` OK
- scuttlebutt `pnpm run typecheck` + vitest 8 passed; `pnpm build` OK

sbluemin:

## Addressed
- P1 Theater change during context collection — `core/client/src/components/quick-launch.tsx` — the invalidation effect now depends on `theaterId` as well as `prompt`, so switching Theater advances the epoch and clears any pending or shown pack.
- P2 Malformed watch output — `terminal/server/agent-api/experiments-routes.ts` — `parseWatchVerdict` returns an explicit `clear` only for `alert:false`; blank, fenced, or invalid answers are recorded and published as `failed` with reason `unreadable_verdict`.
- P2 Watch status after refresh — `terminal/client/agent/experiments-api.ts` (`readWatchLast`) / `agent/index.tsx` — the caption label falls back to the sanitized `payload.watch.last` when no live SSE event exists; the bubble still shows only for live events.
- P2 Snapshot cleanup — `scuttlebutt/server/chat-routes.ts` / `session-registry.ts` (`liveIds`) — snapshots are pruned to live session ids on every `/start` and dropped when a message hits `not_found`, so the map is bounded by the concurrent session cap.

## Validation
- fleet-console typecheck + vitest 332 passed; build OK
- terminal typecheck + vitest 154 passed; build OK
- scuttlebutt typecheck + vitest 8 passed; build OK

sbluemin:
Third-pass feedback has been dispositioned; this closes the review loop for this PR (three passes).

## Addressed
- P2 Embedded settings routes — `core/client/src/settings/settings-pane.tsx` — a requested section that lives inside another chip (`connectivity`, experiment-group plugin sections) now resolves to its host chip, so the remote-access picker and route adapter open the Experiments chip instead of falling back to Appearance.
- P2 Stale payload on watch toggle — `terminal/server/agent-api/experiments-routes.ts` — the Operation is re-read after the body await and immediately before the patch.
- P2 Refinement on Theater change — `core/client/src/components/quick-launch.tsx` — `theaterId` joins the refinement invalidation effect, matching the context pack.

## Declined
- P2 Wrapped watch verdicts: extracting the JSON object from a fenced answer is intended. A fenced `{"alert":false}` is the model's explicit verdict, not a false clear; the prompt asks for JSON only and models routinely fence it. `unreadable_verdict` covers answers with no parseable object or no boolean `alert`, which is the false-assurance case.
- P2 Composite selection key: the two shipped providers use disjoint id spaces (Wiki entry ids vs commit hashes), so the collision is not reachable on a supported path. Not changed.

## Validation
- fleet-console typecheck + vitest 332 passed; build OK
- terminal typecheck + vitest 154 passed; build OK

PR #1011, opened 2026-09-06
ci(mobile): retry transient App Store Connect failures

sbluemin:
## Summary
- The v1.85.0 Stable Release TestFlight job uploaded the IPA (`Upload accepted`, build 0.3.1 (4)) and then died with an unhandled `ConnectTimeoutError` while polling App Store Connect for processing to finish — one dropped connection ended a distribution Apple had already accepted.
- Every ASC request now retries with exponential backoff (4 attempts, 2s/4s/8s) on transport failures and 429/5xx, while permanent rejections (401/403/404/409) still fail on the first response.
- Applied at the request layer rather than only in the poll loop, so the tester assignment, release notes, and beta review calls are covered too; those create-style calls already recover from duplicates.

## Test Plan
- [x] `pnpm build`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `runtime/fleet-mobile` — new `App Store Connect transport` cases: dropped connection is retried then succeeds, exhausted retries fail naming the transport cause, a permanent 401 is not retried

### DIFF
--- a/runtime/fleet-mobile/scripts/lib/asc-api.mjs
+++ b/runtime/fleet-mobile/scripts/lib/asc-api.mjs
@@ -20,11 +20,24 @@ export const TOKEN_LIFETIME_SECONDS = 900;
 // 오류와 Apple의 혼잡 응답은 재시도로 흡수하고, 영구적인 거절만 호출자에게 올린다.
 export const RETRY_ATTEMPTS = 4;
 export const RETRY_BASE_DELAY_MS = 2000;
+// Apple이 Retry-After로 지정한 대기가 이보다 길면 재시도해도 백오프 안에서 같은 창에 갇힌다.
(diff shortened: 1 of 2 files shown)


sbluemin:
@codex Please review this PR.

sbluemin:
**FIX** — `asc-api.mjs` Retry-After (P2): applied in `f193a1bf4`.

Error responses now carry a parsed `Retry-After` (seconds or HTTP-date), and the next attempt waits at least that long instead of the fixed 2s/4s/8s backoff. When Apple asks for longer than this job can usefully retry (60s), the request stops on the first response and reports the requested wait rather than burning the remaining attempts inside the same rate-limit window.

Covered by two new cases in `runtime/fleet-mobile/tests/ios-tools.test.ts`: a `Retry-After: 30` response sleeps 30s then succeeds; `Retry-After: 3600` fails after a single call.

@codex Please review this PR.

PR #1014, opened 2026-09-06
fix(console): keep Quick Launch Run on one row and move attach to the prompt

sbluemin:
## Summary
- Quick Launch control row no longer wraps at desktop widths: the row is `nowrap`, and the Theater/model chips shrink with ellipsis (floors 92/108px) instead of the Run button dropping to a second line. Root cause: flex line-breaking uses pre-shrink max-content widths, so the selection's `min-width: 0` never took effect (578px demand vs 556px budget with a 14-char model label and the ULTRACODE gate open). Mobile (<=767px) keeps the wrapping layout.
- The image-attach control moves from the control row to the top-right of the prompt input beside the refine trigger, and the mention/dock toggles form one tight group. The selection budget grows 556 -> 600px, so the reproduced combination fits with no truncation; longer labels fall back to the one-row contract.
- Overflow/rejection status text shrinks with ellipsis rather than forcing a wrap.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console typecheck`, `vitest tests/quick-launch.test.ts`, `tests/instrument-design-contract.test.ts`
- [x] `pnpm --filter @dotobokuri/fleet-console build` and isolated Console e2e (1440x900): reproduced combination stays one row (bar 55px, no truncation); 24-char model label truncates both chips and stays one row; pinned dock expand/collapse; 375px keeps wrap with track fully visible; refine trigger + attach co-placement (`right: 30px`, input `padding-right: 62px`)

### DIFF
--- a/.changelog.d/quick-launch-bar-fit-0906.md
+++ b/.changelog.d/quick-launch-bar-fit-0906.md
@@ -1,0 +1,11 @@
+---
+branch: quick-launch-bar-fit-0906
+---
+
+### fleet-console
+#### Fixed
+- Quick Launch keeps the Run button on the control row even when a long model name or the open ULTRACODE gate fills the row; Theater and model chip labels shorten instead of pushing Run onto a second line.
+  ko: Quick Launch에서 긴 모델 이름이나 열린 ULTRACODE 게이트가 줄을 채워도 실행 버튼이 컨트롤 행에 남습니다. 실행 버튼이 둘째 줄로 밀리는 대신 Theater·모델 칩 라벨이 줄어듭니다.
+#### Changed
+- The Quick Launch image-attach button moved from the control row to the top-right of the prompt input, next to the prompt refine button, and the mention and dock toggles now sit as one tighter group beside Run.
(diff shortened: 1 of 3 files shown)


sbluemin:

## Addressed
- Highlight mirror alignment — `components.css` (`.quick-launch-input-wrap:has(...) .quick-launch-highlight`) — the mirror takes the same 32px/62px right padding as the textarea when attach and/or the refine trigger are present. Reproduced before the fix: with attach present and an `ultracode` prompt wrapping at the right edge, mirror text ran under the attach icon and the caret sat ~125px away from the mirrored text end; after the fix both layers wrap at the same width (mirror max right 1053px vs textarea content right 1051px).
- Mention target tag overflow — `components.css` (`.quick-launch-target-tag { min-width: 0 }`, new `.quick-launch-target-label` with ellipsis) and `quick-launch.tsx` (label span gets the class and the full Theater name as `title`). Reproduced with a 96-char Theater folder name and a dormant Operation: the tag spilled 220px past the card's left edge; after the fix the tag ellipsizes at 592px and Run/toggles stay in place.

## Validation
- `vitest tests/quick-launch.test.ts tests/instrument-design-contract.test.ts` — 100 passed
- Isolated Console e2e (1440x900 @2x) on the rebuilt bundle for both scenarios above; no console errors.

## Notes
- Before the fix the overflow pushed the tag out to the left (bar has `justify-content: flex-end`), not Run out to the right as described, but the defect and its fix are the same.

PR #1015, opened 2026-09-06
feat(codex): rebuild Cowork as a revision thread on the Console composer grammar

sbluemin:
## Summary
- Rebuild the Codex Cowork dock as a revision thread using the Agent Chat grammar: turn spine, live tally with thinking/tool steps, "worked for" fold, streaming answer, review dock (changes/full), and per-cause notices (turn failure, timeout, stale draft, gateway down, disabled model).
- Assemble the composer from `@fleet-console/sdk/composer` blocks with a provider-glyph model chip and banded portal menu (Analyst composition), plus an effort track; the input stays focusable during a turn so Esc stops it (the expanded reading sheet yields Escape to a running composer).
- Add Codex GPT-5.6 Luna to the Cowork model list when it is enabled in Settings › AI Gateway; the codex plugin now reads the Fleet-root gateway selection so disabled models are never offered. Drydock proposers render as readable provenance marks; cowork CSS is rewritten on host tokens only with dead motions/selectors removed.

## Test Plan
- [x] `@fleet-plugins/codex` typecheck + vitest (21 passed)
- [x] `@dotobokuri/fleet-console` typecheck, build, `instrument-design-contract` and codex route tests (108 passed)
- [x] `check:localized-attributes`, `check:core-boundary`, changelog fragment `--check`
- [x] console-e2e (isolated Console, headless): empty hero → model menu shows Claude/Codex bands with Luna → live Luna turn (thinking, `ToolSearch`/`wiki_draft_read` steps, streaming answer, review dock, changes diff) → second turn in the same thread → Apply → published body updated and Drydock archive shows the `codex-cowork` proposer mark → Esc stops a running turn without closing the sheet; 0 page errors/rejections
- [ ] Not verified: Whites theme rendering of the thread; split (non-expanded) reader pane Escape handling


sbluemin:

## Addressed
- Match notices to emitted error codes — `packages/fleet-wiki/src/cowork/service.ts:108`, `runtime/fleet-plugins/codex/server/codex/cowork/gateway-adapter.ts:132`, `cowork-controller.ts:52` — the service now forwards connector `cowork_*` codes (timeout/incomplete/model-not-enabled) and the client maps `provider_unavailable`/`provider_error`/`cowork_model_not_enabled` to the gateway/turn/disabled-model notices.
- Yield Escape when the model menu is open — `codex-reading-sheet.tsx:265` — the sheet skips Escape whose target is inside `.cowork-model-menu` or the expanded chip, so the menu closes itself and focus returns to the chip.
- Keep the composer outside the thread scroll viewport — `cowork-thread.tsx`, `styles/components.css` — only `.cowork-thread-scroll` (turns/notices/panel) scrolls; the review dock and composer stay fixed below, and the history pins to its end while streaming unless the user scrolled up.

## Validation
- `pnpm --filter @dotobokuri/fleet-wiki build && test` — 39 passed
- `@fleet-plugins/codex` typecheck + vitest — 21 passed
- `@dotobokuri/fleet-console` build; `instrument-design-contract` — 97 passed
- console-e2e (isolated, headless, rebuilt bundle `index-gwWU7WHF.js`): model menu open → Escape closes only the menu, focus returns to the chip, sheet stays; composer/review dock render outside `.cowork-thread-scroll` (overflow-y auto, max 420px) with the root non-scrolling

sbluemin:

## Addressed
- Create a turn when an SSE run starts — `cowork-controller.ts:488` — an observed idle→running `session` event now appends a running placeholder turn (labeled "Instruction sent from another window"), so transcript/tool events land, `isRunning()` is true, and Stop is shown; same contract as the mount-time restore path.

## Validation
- `@fleet-plugins/codex` typecheck + vitest — 21 passed
- Console client bundle rebuilt (vite) — OK
- Reproduced by code path: `isRunning()` required `currentTurn()` and `transcript` returned early without a turn; the placeholder restores the pre-rewrite behavior where a foreign run was visible.

sbluemin:
Addressed the third-pass finding and stopping the review loop here (three passes; each FIX was reproduced from the code path).

## Addressed
- Ignore stale idle snapshots when restoring a live turn — `cowork-controller.ts` — a replayed `session` snapshot whose state is `running` now clears any pending settlement timer, so the mount-time replay (idle → running) no longer marks the still-active turn as stopped.

## Validation
- `@fleet-plugins/codex` typecheck + vitest — 21 passed; Console client bundle rebuilt

PR #1036, opened 2026-09-08
fix(ai-gateway): drop tool regex patterns the Responses backend rejects

sbluemin:
## Summary
- The ChatGPT Codex backend compiles every function tool's `pattern` with RE2 and refuses the whole request (`param: "tools"`) when it cannot read one. Claude Code's `Artifact` tool carries five such patterns (lookahead in `collection`/`doc_id`/`writes.*`, lookahead plus `\p{...}` in `field`), so every Console Operation on a Codex (GPT) gateway model died with `400 Invalid schema for function 'Artifact'` on its first message.
- Sanitize every tool's schema ahead of the strict branch and drop only the patterns RE2 rejects, keeping the ones it accepts. The refusal is independent of `strict`, so the old strict-only path never ran for `Artifact` at all. `pattern` only advises the model about a value's shape, which is the same trade `strictSchema` already makes for `format`.
- Apply the same rule to the opencode-go Responses adapter, which serves GPT models (`opencode--gpt-5.6-luna`) over the identical wire contract. That backend is not live-verified here; the change is inert on a backend that would have accepted the patterns.

## Test Plan
- [x] `pnpm --filter @dotobokuri/core-ai-gateway test` (28 files, 193 tests) with a new adapter case asserting the hostile patterns are dropped and RE2-safe ones survive; verified the case fails on the unpatched adapter.
- [x] `pnpm typecheck`, `pnpm test` (full workspace), `pnpm check:core-boundary`, `pnpm check:localized-attributes`, `pnpm check:claude-agent-sdk-boundary`, `node --test scripts/*.test.mjs`, `node scripts/compile-changelog-fragments.mjs --check`.
- [x] Live: the real `Artifact` schema through an unpatched wire returns `400 ... is not a 'regex'`; the same schema through the patched gateway to the live Codex backend returns 200 and answers, with only the three RE2-safe patterns reaching the upstream body.

### DIFF
--- a/.changelog.d/codex-tool-pattern-re2.md
+++ b/.changelog.d/codex-tool-pattern-re2.md
@@ -1,0 +1,13 @@
+---
+branch: codex-tool-pattern-re2
+---
+
+### fleet-cli
+#### Fixed
+- Claude Code sessions on Codex (GPT) gateway models answer again. A tool whose JSON Schema carried a regular expression the Responses backend cannot compile, such as the lookahead and Unicode property patterns in Claude Code's Artifact tool, failed the entire request with a 400 before any reply arrived.
+  ko: Codex(GPT) 게이트웨이 모델의 Claude Code 세션이 다시 답변합니다. Claude Code Artifact 도구의 lookahead·유니코드 속성 패턴처럼 Responses 백엔드가 컴파일하지 못하는 정규식이 도구 JSON Schema에 있으면, 답변이 오기 전에 요청 전체가 400으로 실패했습니다.
+
+### fleet-console
+#### Fixed
+- Chat in an Operation running on a Codex (GPT) gateway model works again. A tool whose JSON Schema carried a regular expression the Responses backend cannot compile, such as the lookahead and Unicode property patterns in Claude Code's Artifact tool, failed the entire request with a 400, so the first message you sent never produced an answer.
+  ko: Codex(GPT) 게이트웨이 모델로 도는 Operation의 채팅이 다시 동작합니다. Claude Code Artifact 도구의 lookahead·유니코드 속성 패턴처럼 Responses 백엔드가 컴파일하지 못하는 정규식이 도구 JSON Schema에 있으면 요청 전체가 400으로 실패해, 처음 보낸 메시지가 끝내 답변을 받지 못했습니다.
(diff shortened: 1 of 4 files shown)


sbluemin:
@codex Thanks — reviewed and **declined**, with evidence.

## P2 "Avoid dropping valid patterns containing lookaround text" — DECLINE

**Not reproducible on a supported path.** The false positive needs a character class whose literal contents are `(?=`, `(?!`, or `(?<` in that order (your `^[(?=]+$`). Scanning every `pattern` that real Claude Code tool catalogs actually put on the wire (captured from live gateway runs) gives 4 distinct patterns, and the detector keeps all 4:

```
KEEP | Artifact | ^(0|[1-9]\d{0,3})\.(0|[1-9]\d{0,4})\.(0|[1-9]\d{0,5})$
KEEP | Workflow | ^wf_[a-z0-9-]{6,}$
KEEP | Artifact | ^[A-Za-z0-9_=-]{1,4096}$
KEEP | Artifact | ^[0-9a-f]{32}$
```

On the full `Artifact` schema it drops exactly the 5 real lookaround/`\p{...}` patterns and keeps the 3 safe ones — verified live against the Codex backend (200, answer returned; the same schema unsanitized returns `400 ... is not a 'regex'`). No observed false positive.

**The stated consequence does not hold here.** `pattern` is advisory: this gateway never validates tool arguments against it, and the tool handler owns its own validation. Dropping one is the same trade `strictSchema` already makes for `format` (see the comment above it). For a non-strict tool — which is where `Artifact` lands — the backend does not enforce `pattern` at all, so nothing downstream weakens.

**The proposed remedy has the worse failure mode.** "Detect unsupported constructs in regex syntax" means a character-class/escape-aware scanner. Its false negatives reintroduce the defect this PR fixes: a request-wide `400` (`param: "tools"`) that kills the turn before any output, which is how every Console Operation on a Codex model died at its first message. Over-dropping an advisory constraint costs a generation hint; under-detecting costs the whole session. The bias is deliberate, and it mirrors the same asymmetry the neighbouring `STRICT_ALLOWED_KEYWORDS` allowlist is built on.

If a real tool ever ships such a character class, that is the point to pay for a scanner.

PR #1039, opened 2026-09-09
feat(repository): refit the panel with a tree-led context, one work bar, and shared glyphs

sbluemin:
## Summary
- Remove the Repository top context band: the repository/worktree switcher leads the sidebar tree, and Fetch/Pull/Push/Stash share one work bar with the History/Changes segmented tabs and the portaled history toolbar, collapsing labels -> icons -> a More menu as the panel narrows.
- Restructure the tree into Worktrees / Branches / Remotes (grouped per remote with a host mark) / Tags / Stashes with a check-marked current branch and ahead/behind gauges (`for-each-ref upstream:track`), replace unicode glyphs and underline tabs with a 16-grid SVG set and three control tiers, and swap the always-on status bar for a transient toast.
- Make list-view changed-file rows single-line with a leading ellipsis, and rebuild the commit File Tree tab in the File Explorer grammar with a content pane (diff for changed files, blob content via a new `blob` route otherwise); the file icon module moves to `@fleet-console/sdk/components/file-icon` so both plugins share it.

## Test Plan
- [x] `tsc --noEmit` for repository, file-explorer, and fleet-console (host + client)
- [x] `vitest run` repository (76) and file-explorer (31)
- [x] `check-localized-attributes`, changelog fragment `--check`
- [x] Isolated headless Console (console-e2e, 4 runs): identity band gone, work bar 36px with segmented tabs, progressive collapse at work-area 977/747/547/427/395 with no overflow, More-menu Escape keeps the surface, picker rows 30px with fixed sizing, tree sections with GitHub mark and ahead/behind, dock header icons, list rows 26px with leading ellipsis, File Tree tab with blob/diff pane, toast on Fetch, light theme; 0 page errors
- [ ] Not verified: stash popover from the More-menu path, upstream `gone` state, viewport <= 800px (Console switches to mobile layout)


sbluemin:

## Addressed
- P2 Bound blob lines — `client/commit-tree.tsx` — the content pane renders at most 5000 lines (`BLOB_MAX_LINES`) and marks the preview truncated; the 1 MB byte cap on the route stays.
- P2 Reset tree selection — `client/history-panel.tsx:347` — the commit-loading effect now clears `treeSelection` alongside the loading state, so parent/child navigation never requests the previous commit's path.

## Validation
- `tsc --noEmit` (repository) — clean
- `vitest run` (repository) — 76 passed

sbluemin:

## Addressed
- P2 Dash-prefixed blob paths — `server/blob.ts` — replaced the directory predicate with `isSafeBlobPath()` (rejects NUL, absolute paths, and `..` traversal only); leading `-`/`:` names are accepted since the path is appended to the validated hash as one `<sha>:<path>` argument.

## Validation
- `tsc --noEmit` (repository) — clean
- `vitest run` (repository) — 76 passed

sbluemin:
Review pass 3 addressed. Dispositions below; stopping the review loop here per the three-pass rule (no FIX-class finding is left open).

## Addressed
- P2 Reset folder caches when switching commits — `client/history-panel.tsx` — `CommitTreeView` is keyed by `target.fullHash`, so folder cache and expansion state are rebuilt per commit.
- P2 Match remote group names in ref searches — `client/rail-panel.tsx` — a query matching the remote name or host keeps the whole group, and rows also match the qualified `remote/branch` label.

## Deferred
- P2 Stop reading blob output at the preview limit — `server/blob.ts` — the drain-after-cap behavior belongs to the shared `runGit` executor and applies equally to the existing diff/commit-file routes; it is bounded by the executor timeout. An abort-at-cap in `git-executor.ts` is a cross-route change outside this PR and should land as its own change.

## Validation
- `tsc --noEmit` (repository) — clean; `vitest run` — 76 passed; `pnpm --filter @dotobokuri/fleet-console build` — built

PR #1042, opened 2026-09-11
fix(repository): keep current content on screen while reloading

sbluemin:
## Summary
- Every refetch in the Repository panel (refresh, sync, commit, staging, branch selection, commit navigation) used to clear its surface to an empty/loading state first, so the branch tree, history list, changed-file count, WIP row, toolbar buttons, and inspector bodies flashed on each round-trip. All of them now keep the last result until the next response lands (stale-while-revalidate), and only the refresh glyph spins while a load is in flight.
- Same-checkout reselection no longer remounts the History panel; a `landingSeq` prop resets filter/selection/compare/scroll in place and refetches. Same-slot refreshes keep scroll position; ref/order changes jump to the top only when the new list arrives.
- Commit/compare/stash inspectors, hunk diff, and blob view keep their previous body during a target switch and dim it after a 160ms delay, so fast local git round-trips paint no change at all.

## Test Plan
- [x] `@fleet-plugins/repository` typecheck, build, and vitest (15 files / 76 tests)
- [x] `@dotobokuri/fleet-console` build; `instrument-design-contract` test (98)
- [x] Isolated headless Console (console-e2e) with a MutationObserver sampling DOM state: tree reload fires 7 requests with zero visible DOM changes and scroll preserved; branch selection swaps 42 → 35 rows with no empty state; same-repo reselect clears the ref chip without a remount; commit navigation shows no loading text (stale dim only); staging keeps the Changes count and list rows while the file moves to Staged; 0 page errors/rejections


sbluemin:

## Addressed
- P1 one-shot sequences — `history-panel.tsx` landing branch — reset `handledCompareRequestSeqRef` / `handledInspectRequestSeqRef` / `handledStashRequestSeqRef` to 0 when `landingSeq` changes, so a compare/stash/inspect request that restarts at seq 1 is consumed again.
- P2 blob scroll — `commit-tree.tsx` `CommitBlobView` — added a scroll ref and a layout effect that scrolls to the top when the replacement content lands (same pattern as `HunkView`).
- P2 worktree spinner — `rail-panel.tsx` — added an explicit `worktreesPending` flag cleared in `finally`; the reload glyph no longer keys off `worktreesForRepoRel !== repoRel`, which stays null after a failed lookup by design (restore-effect gate).
- P2 stuck history reloading — `history-panel.tsx` cache-restore branch — `setReloading(false)` when a cached slot fills the list while the previous fetch was cancelled.

## Validation
- `pnpm --filter @fleet-plugins/repository typecheck` / `build` / `test` — pass (15 files, 76 tests)

sbluemin:

## Addressed
- P2 commit identity — `history-panel.tsx` `CommitInspector` — the ok state now carries `fullHash`; `commit`, `CommitTreeView`, `CommitBlobView`, and `CommitHeader` derive from that loaded hash while the previous body is retained, so `commit-file` is never requested with the new hash and an old path.
- P2 compare identity — `compare-inspector.tsx` — the ok state now carries its `pair`; `compareSelection` derives from the loaded pair while the previous file list is retained, so `compare-file` is never requested for a new range with a file from the old list.

## Validation
- `pnpm --filter @fleet-plugins/repository typecheck` / `build` / `test` — pass (15 files, 76 tests)

sbluemin:
The third-pass finding has been addressed in `6fb335e91`.

## Addressed
- P2 pagination during slot switch — `history-panel.tsx` — `loadMore` bails and the "Load more" button is disabled while `reloading`, so the previous list's length is never sent as `skip` against the new slot's query.

## Validation
- `pnpm --filter @fleet-plugins/repository typecheck` / `build` / `test` — pass (15 files, 76 tests)

This closes the third review pass; proceeding to the final context audit and merge.

PR #1059, opened 2026-09-12
feat(console): flatten agent turns and animate the thinking tail

sbluemin:
## Summary
- Remove the left spine and status node from agent turns in Chat view, the Analyst panel, and the Codex cowork thread; the working clock, fold line, and outcome text already carry that state, so each turn is now a single column.
- Hide the thinking dots with `visibility` instead of `opacity`: the live-text background clip painted the parent gradient through the hidden dots' glyph shapes, so the `.` → `..` → `...` cycle never showed.
- Spin the dashed thinking glyph slowly (3s) while the live tail says thinking.

## Test Plan
- [x] `pnpm typecheck` in `runtime/fleet-console` and `runtime/fleet-plugins/codex`
- [x] `vitest run tests/instrument-design-contract.test.ts core/client/src/agent` (fleet-console) and codex plugin tests
- [x] Isolated Console built from this branch: live Gemini Flash chat turn shows no spine, single-column turn, dots cycling (`vvv` → `vvh` → `vhh` visibility samples), rotating think glyph; settled turn renders fold + answer; no page errors
- [ ] Analyst panel and Codex cowork thread verified visually (CSS/typecheck only)

### DIFF
--- a/.changelog.d/chat-stream-flat.md
+++ b/.changelog.d/chat-stream-flat.md
@@ -1,0 +1,10 @@
+---
+branch: chat-stream-flat
+---
+
+### fleet-console
+#### Changed
+- Flatten agent turns in Chat view, the Analyst panel, and the Codex cowork thread by removing the left spine and status node; the working clock, fold line, and outcome text already carry that state.
+  ko: 채팅뷰, Analyst 패널, Codex cowork 스레드의 에이전트 턴에서 좌측 세로 스파인과 상태 노드를 제거해 평평하게 만들었습니다. 진행 시계, 접힘 줄, 결말 문구가 이미 그 상태를 말합니다.
+- Animate the thinking tail in Chat view: the dots now step through one, two, and three, and the dashed thinking glyph slowly spins while the model is between tool calls.
+  ko: 채팅뷰의 생각 중 꼬리에 움직임을 더했습니다. 점이 하나, 둘, 셋으로 차오르며 순환하고, 도구 호출 사이에 점선 생각 글리프가 천천히 회전합니다.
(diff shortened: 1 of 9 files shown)


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed
- P1 (cowork error indicator) — `runtime/fleet-plugins/codex/client/codex/cowork-controller.ts` send() catch — set `notice = noticeFrom(cause)` so a pre-stream failure surfaces the notice card (alert + retry for turn-class codes). Only `ensureSession` actually lacked a notice: `updateCoworkAnnotations` and `promptCowork` already go through `mutate`, which sets the same notice before rethrowing, so the assignment is idempotent on those paths.

## Validation
- `pnpm --filter @fleet-plugins/codex typecheck` — pass
- `vitest run` (codex plugin) — 32 passed

## Notes
- No new permanent test: the send-failure path has no existing controller harness, and the fix is a one-line notice assignment on an already-covered notice surface.

sbluemin:

## Addressed
- P2 (reduced-motion spin) — `runtime/fleet-console/core/client/src/agent/chat/chat.css` reduced-motion block — the override now uses the same selector as the live spin rule (`.agent-chat-tally.is-live .agent-chat-tally-glyph > .agent-glyph[data-glyph="think"]`), so it wins on equal specificity by source order and the rotation stops under `prefers-reduced-motion: reduce`.

## Validation
- `vitest run tests/instrument-design-contract.test.ts` (fleet-console) — 98 passed
- `vite build` (Console client) — pass

## Notes
- None declined or deferred.

PR #1079, opened 2026-09-13
feat(console): move AI extension switches into the Operation menu and gate Computer Use per Operation

sbluemin:
## Summary
- Session watch, Console use and Computer Use leave the caption band and stand as one "AI extensions" section at the top of the shared Operation menu (caption ···, sidebar right-click, War Room card). Rows are one-line menu items with state in the tooltip; active extensions also show as 12px glyph marks on the sidebar chip. The Session watch result bubble stays on the caption, anchored under the ··· button.
- Computer Use adopts the Console use policy: tools ride every session, every fleet-computer-use call is refused at call time until the caller Operation is allowed from its menu, allowing/revoking applies to the next call, a refusal names the switch and where to turn it on, and revoking releases the device session that Operation held.
- The Console use caption glyph is redrawn as a window whose corner a pointer takes; Computer Use (screen + stand) and Browser Use (window + address bar) marks share the pointer. SDK gains `operationMenu` and `operationMarks` hooks.

## Test Plan
- [x] `pnpm typecheck` / `pnpm build` in runtime/fleet-console
- [x] fleet-console vitest, including a new per-Operation Computer Use authorization/revoke case
- [x] Isolated Console: menu rows from caption ··· and sidebar right-click, toggling writes/clears `payload.computerUse`, chip marks appear/disappear, caption shelf keeps only Analyst/view controls, no page errors
- [ ] Live Session watch review to confirm the bubble position under ··· (needs a real turn)


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed
- P1 (computer-use.ts:87, revoke misses a call still in preflight) — FIX in `core/host/mcp/computer-use.ts`: the MCP host tracks in-flight calls per caller Operation and `revokeOperation` aborts them before releasing the owner; the service's pre-dispatch `signal.aborted` check then refuses the call before it reaches the native broker. Reproduced with a new test that parks `resolveTarget` on a gate, revokes mid-await, and asserts `computer_use_session_unavailable` with the broker never created.

## Validation
- `pnpm typecheck`, `pnpm build` (runtime/fleet-console) — pass
- `vitest run tests/execution/computer-use.test.ts` — 10 passed

PR #1081, opened 2026-09-13
feat(console): move the Operation folder from the sidebar chip into a hover card

sbluemin:
## Summary
- With "Show Operation location" on, the sidebar Operation chip now carries only the git branch; the working folder moves into a compact card that opens on hover or focus.
- The card holds three rows and nothing else — status, current location, start time. No title, no execution details, no location glyphs or Theater root/outside wording, no operating hints.
- The chip's native `title` steps aside while the card is available so the two tooltips cannot overlap; turning the experiment off restores today's chip and tooltip exactly.

Direction was chosen from a measured proposal: in the worktree convention the branch and the Operation title are the same string, so the chip was truncating the same text twice while the folder had no room. The Cmd+K palette row is deliberately unchanged — it is wide and is not a list people scan.

## Test Plan
- [x] `pnpm typecheck` (all workspaces)
- [x] `pnpm --filter @dotobokuri/fleet-console exec vitest run` — 525 passed, 8 skipped
- [x] `pnpm check:localized-attributes`, `pnpm check:core-boundary`
- [x] Isolated Console, experiment on: chip shows the branch only and reports `title=null`; card opens on hover (dwell) and on keyboard focus, closes on Escape/leave/scroll/resize; page errors 0
- [x] Seeded a chat Operation that moved its cwd into `docs` — session DTO reports `folder: "docs"`, and the card's location row reads `feature/tooltip-demo · docs` while the chip stays branch-only
- [x] Experiment off: chip returns to a single line and the native tooltip comes back


sbluemin:

## Addressed
- Keep open detail cards synchronized — `operations-side-bar-chip.tsx` / `operation-detail-card.tsx`: state now holds only the chip rect; activity, location and start time are passed live, and the placement effect re-measures when the content changes. Reproduced: with the card held open, a turn-start hook flipped the status row from "유휴" to "실행 중" in place.
- Close the card when an interaction becomes blocked — `operations-side-bar-chip.tsx`: `detailBlocked` is a value, an effect dismisses the open card and cancels the pending timer when it turns true, and the hover timer re-reads the current state through a ref instead of its captured render. Reproduced: an open card survived arming close before the fix and disappears after it.
- Associate the detail tooltip with its chip — the portaled card takes a `useId()` id and the chip sets `aria-describedby` to it while open. Verified in the browser: `aria-describedby` equals the rendered card's id on both the pointer and keyboard paths.
- Update the setting copy for the moved folder — `i18n/messages/pages.ts`: both locales now say the chip carries the branch and the hover card carries status, current location and start time.

## Validation
- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit -p tsconfig.json` — clean
- `pnpm --filter @dotobokuri/fleet-console exec vitest run` — 525 passed, 8 skipped
- Isolated Console with the experiment on: chip branch-only with `title=null`; card opens on hover dwell and on focus, follows a live activity change, closes on Escape/leave/scroll/resize and on arming close; page errors 0

## Notes
- No new permanent tests: these are UI micro-interaction and copy behaviours, which the repository's test admission policy keeps out of the suite. Coverage is the real-app verification above.

sbluemin:

## Addressed
- Preserve folder details for preview chips — `operations-side-bar-chip.tsx`: `preview` no longer blocks the detail card. What a preview chip drops is close/rename/accent/reorder — ways to change an Operation, not ways to read one — and the card is read-only and pointer-transparent, so blocking it there was the one place that lost the folder with no replacement. The native title now steps aside on the same condition as every other chip, so the two tooltips still cannot overlap.

## Validation
- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit -p tsconfig.json` — clean
- `pnpm --filter @dotobokuri/fleet-console exec vitest run` — 525 passed, 8 skipped
- Reproduced in an isolated Console with two Theaters and the experiment on: the inactive Theater's chip renders as a preview chip (no close button), and both the hover and keyboard-focus paths now open the card reading `상태 유휴 / 위치 feature/tooltip-demo / 시작 2분 전`, with `aria-describedby` matching the card id. Page errors 0.

## Notes
- This is the third pass on this PR. Under the repository's review contract I will stop the loop here unless a further pass names a reproduced defect on a supported path.

sbluemin:
Third review pass — partially addressed, with reasons for the rest. Closing the review loop here per the repository's review contract.

## Addressed
- Avoid opening the card over titled descendants (the location line) — `operation-workspace-context.tsx` / `operations-side-bar-chip.tsx`: the branch and folder spans no longer carry a `title` in the sidebar projection. That was the real collision: the branch span is the widest target on the chip's second line, and its tooltip duplicated a value the card already shows in full. The Cmd+K palette has no card, so it keeps its titles. Verified in an isolated Console: the chip's branch span reports `title=null` while the same branch in the palette still reports `브랜치 feature/tooltip-demo`.

## Declined
- The same comment's other targets — the close and minimize buttons, and the Theater and group pills — keep their titles.
  - The buttons' titles name the control under the cursor. They predate this PR, they are the only mouse-visible label those icon buttons have, and the card does not replace them. Suppressing them would remove an affordance nobody asked to remove.
  - The pills' titles reveal a truncated Theater or group name. The card carries neither, so dropping them would delete information with no replacement.
  - A tooltip describing the specific control you are pointing at is not the collision this PR set out to remove; the collision was the chip-wide operating tooltip, and that one is gone.

## Validation
- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit -p tsconfig.json` — clean
- `pnpm --filter @dotobokuri/fleet-console exec vitest run` — 525 passed, 8 skipped
- Isolated Console, experiment on: chip `title=null`, branch span `title=null`, palette branch title intact, page errors 0

PR #1083, opened 2026-09-13
fix(console): make the Operation card's location the session's working place

sbluemin:
## Summary
- The detail card's location row joined the branch first, so an Operation sitting at its Theater root read as `canary` — a branch name standing in for a place.
- The branch is already on the chip, so the card drops it and roots the row at the Theater name instead: `handoff-theater` at the root, `handoff-theater/docs` below it, and `…/<last segment>` for a session that walked outside the Theater. No glyphs, no explanatory wording, no absolute paths.
- Chips subscribe to just the Theater name through a new string-snapshot selector, so resolving it does not re-render the list on unrelated state changes.

No changelog fragment: `.changelog.d/operation-context-tooltip-proposal.md` is still unreleased and already describes the card as carrying "the Operation's status, current location, and start time". This makes the implementation match that description rather than changing behavior users could consume, so per the Release Baseline rule it needs neither a new note nor an amendment.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit -p tsconfig.json`
- [x] `pnpm --filter @dotobokuri/fleet-console exec vitest run` — 527 passed, 8 skipped
- [x] Isolated Console with the experiment on and two seeded Operations in one git Theater: the session that moved into `docs` reads `위치 handoff-theater/docs`, the one at the root reads `위치 handoff-theater`, and neither shows the branch in that row. Page errors 0.
- [x] Cmd+K palette rows are untouched — they keep the branch-and-folder line.


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed
- Pass the unfiltered workspace to the detail card — `operations-side-bar-chip.tsx`: the card now receives `session?.workspace` instead of the `visibleWorkspace` projection. That filter exists for the chip line, which has nothing to draw when there is neither branch nor folder; the card still has something to say, because the session stands in the Theater.

## Reproduced
- Registered a plain folder (no `.git`) as a Theater and seeded one Operation in it. The session DTO reports exactly the shape described: `{"folder":null,"outside":false,"branch":null}`.
- Before the fix the card showed only 상태 and 시작. After it the card reads `상태 유휴 / 위치 plain-theater-nogit / 시작 29초 전`, and the chip is still a single 38px line with no location row. Page errors 0.

## Validation
- `pnpm --filter @dotobokuri/fleet-console exec tsc --noEmit -p tsconfig.json` — clean
- `pnpm --filter @dotobokuri/fleet-console exec vitest run` — 527 passed, 8 skipped

PR #1091, opened 2026-09-14
refactor(console): drop the Theater badge from War Room sidebar rows

sbluemin:
## Summary
- Remove the Theater pill from sidebar chips. The War Room deck already groups Operations by Theater, so the badge repeated a distinction the map makes, and it ate the width long Operation titles needed.
- Keep the Theater in the chip's accessible name: a screen reader has no map to read the grouping from.
- Delete the now-unused `.side-bar-chip-theater-pill` rule.

## Test Plan
- [x] `pnpm typecheck` (runtime/fleet-console)
- [x] `vitest run tests/instrument-design-contract.test.ts` (98 passed)
- [x] Isolated Console, headless browser: two Theaters and four Operations, War Room entered; `.side-bar-chip-theater-pill` count 0, chip text is the title alone, chip `aria-label` still ends with the Theater name, no JS errors or rejections
- [x] Cruise mode re-checked after the change (chips intact, no regression)


sbluemin:
@codex Please review this PR.

sbluemin:
@codex Thanks — reviewed and declined, with the product owner's confirmation.

## Notes
- **DECLINE — P2 "Preserve Theater identity for off-deck shelf rows"** (`operations-side-bar-chip.tsx:359`). The mechanism is real: `canvas.tsx` excludes ended and minimized Operations from the deck, so shelf rows have no card carrying their Theater. But the disposition is a product decision, and it was put to the requester with that mechanism spelled out: the badge is removed across the whole War Room sidebar, shelves included.

  Keeping a badge on shelf rows only would split the grammar inside one list — a badge that appears on some rows reads as carrying its own meaning rather than as a location label, which is the redundancy this PR set out to remove. Shelf rows still identify their Theater through the chip's accessible name (kept deliberately), the hover detail card's location line, and the Theater initials on the narrow rail. Same-titled Operations shelved in two Theaters at once is a narrow corner, and it is not worth reintroducing an inconsistent badge.

No FIX items in this pass.

PR #1098, opened 2026-09-15
feat(console): show the Operation group as a chip in the panel caption

sbluemin:
## Summary
- Replace the caption's group colour dot + neutral name with a pill chip that reuses the sidebar chip grammar (group-colour wash, border, ink), so a grouped panel reads both that it is grouped and which group it belongs to. Inactive panels carry a lighter tier; the chip still yields to the title first at the 320px minimum width.
- Drop the colour dot from the sidebar group header; the header otherwise keeps its shape (wash, arrow, count).
- Update the design contract: group colour is injected only via `--group-mark` mixes inside the chip and never reaches the caption fill. Chip ink mixes use oklab so teal does not drift green on Whites.

## Test Plan
- [x] `vitest run tests/instrument-design-contract.test.ts` (98 passed) and `core/client` (7 files)
- [x] Isolated Console with a seeded group fixture: chip renders in Tactical/Cruise, active/inactive tiers, instrument + whites themes, 320px truncation order (chip before title), no page errors


sbluemin:
@codex Please review this PR.

PR #1100, opened 2026-09-15
feat(console): move Session Analyst controls to a glyph row above the composer

sbluemin:
## Summary
- Drop the Session Analyst companion caption band: the analyst registers `hideCaption`, the canvas reclaims the caption outset for headless companions (`reclaimCaptionOutset`), and the frame draws all four corners itself.
- Move identity, state, model, reset, artifacts, back-to-chat, export, and clear into one footing row of 24px glyphs above the composer; the composer stays in both chat and artifacts modes, and the artifacts header folds into a title picker on that row.
- Reset and clear arm on the first press and fire on the second (Enter key-repeat cannot confirm); Escape returns from artifacts to chat before closing the panel.

## Test Plan
- [x] `vitest run core/client/src/agent/analysis tests/instrument-design-contract.test.ts` (111 passed) and the full Console vitest run (one unrelated pre-existing failure in `published-manifest.test.ts`: `@modelcontextprotocol/sdk` external not found)
- [x] Console build + isolated headless Console: headless companion frame starts at the operation caption top (+32px body), footing glyphs in both modes, artifact badge pulse, two-press arm/disarm without DELETE, export menu opens upward, Escape path, no page errors
- [x] Handoff Console reviewed by the user


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed
- P1 — `analysis.css` `.session-analyst__footing-lead` — dropped `overflow: hidden`; label truncation stays on the picker's `strong`. Reproduced in a minimal page: with the old rule the first listbox option was not hit-testable (`elementFromPoint` missed it); with the fix it is.
- P2 — `analysis-glyph-button.tsx` / `analysis-artifacts-panel.tsx` — glyph buttons accept a `popup` descriptor; the export trigger now carries `aria-haspopup="menu"`, `aria-expanded`, and `aria-controls={exportId}` instead of `aria-pressed`, and `[aria-expanded="true"]` shares the pressed face.
## Validation
- `vitest run core/client/src/agent/analysis tests/instrument-design-contract.test.ts` — 111 passed

PR #1105, opened 2026-09-16
feat(console): resume a dormant Operation when its stowed panel is opened

sbluemin:
## Summary
- Opening a panel from the minimized shelf now resumes its session when that session is dormant, so the user no longer lands on a Resume card that needs a second click.
- The auto-resume is gated twice: only when the panel actually came out of the minimized shelf, and only when the shared activity axis reads the Operation as ended — focusing a dormant panel that is already on the canvas still leaves it alone.
- Every single-panel open path (standby chip, search, palette, rail tile, formation view) goes through the same restore-and-resume step; the dormant shelf's resume-in-place contract and the bulk "Open all" path are untouched.

## Test Plan
- [x] `pnpm typecheck`
- [x] `pnpm test` (fleet-console 537 passed / 8 skipped)
- [x] `pnpm check:core-boundary`, `pnpm --filter @dotobokuri/fleet-console boundary:guard`
- [x] New integration case in `operations-boot-minimization.integration.test.ts`, confirmed red without the fix
- [x] Console E2E on an isolated instance with a boot-restored dormant fixture: opening the stowed panel fired exactly one `POST /api/v1/agent/sessions/op-dormant/resume {"fresh":true}` and the panel came up live; re-opening the same panel while live fired no resume; no console errors or unhandled rejections


sbluemin:
Both findings reproduced and fixed in 69d5c5de.

**FIX — Resume dormant panels restored in Formation view.** Confirmed: `openInFormation` computed `wasMinimized` but never called the resume hook. This was my own regression — while checking that the new test failed without the fix, a string edit matched the identical closing lines of both callbacks and removed one permanently.

**FIX — Apply auto-resume when retargeting active focus layers.** Confirmed: `setMaximizedOperationId` / `setCompanionOperationId` drop the target from `minimized` through `setFocusLayer`, so selecting a minimized dormant Operation while a maximized or companion layer is active opened it onto its Resume card. Both branches now capture the pre-open minimized state and run the same check; the companion branch reuses the `operationWasMinimized` already recorded before its `await`.

The threaded callback is now `resumeIfDormant`, with each branch owning its own minimized check, since canvas restore and focus-layer promotion take a panel out of the shelf by different means.

Verified on an isolated Console (3 boot-restored dormant fixtures), each branch resuming exactly once on open:
- standby chip -> `POST /api/v1/agent/sessions/op-standby/resume`
- Tactical/formation -> `POST /api/v1/agent/sessions/op-formation/resume`
- maximized-layer retarget -> `POST /api/v1/agent/sessions/op-max/resume`, opening maximized and live

Inverse held: focusing an already-open live panel fired no resume. No console errors or unhandled rejections. The companion-retarget branch is covered by reasoning and the shared helper, not by a separate E2E run.

`pnpm typecheck` and the fleet-console suite (537 passed / 8 skipped) are green; removing either new call individually turns the added integration assertions red.

sbluemin:
@codex Please review this PR.

sbluemin:
**FIX — Retry auto-resume after runtime hydration.** Reproduced and fixed in d608c4f1.

I first suspected this window was unreachable, then found `agent/index.tsx` declares `setHydration("pending")` at install and only flips to `"ready"` when the event stream connects. Reproduced it in the real app by delaying `/api/v1/agent/events` by 6s: the standby chip was clicked at t=310ms, and the panel stayed on its Resume card with no resume request at all, even 12s later.

The path matters more than the narrow timing suggests — boot is exactly when every panel is stowed, so this is the feature's primary moment, not an edge.

The fix keeps the existing safety rule rather than relaxing it. `resumeOperationInPlace` refuses to act while the axis is unobserved, and that stays: the open gesture is held instead of executed, then settled once the axis is authoritative and re-judged against the observed snapshot. It is dropped on `degraded` ("we don't know" must not be promoted to a fact) and dropped if the user stows the panel again while waiting; a closed or actually-live Operation is filtered by the same gate that runs on the immediate path.

Re-verified with the same delayed-stream scenario: click at t=310ms, resume at t=6338ms — a 6028ms gap matching the injected delay — panel live, no dormant card, no errors or unhandled rejections. The re-stow guard is covered by reasoning and the shared gate, not by its own run.

Also confirmed on request that Cmd+K resumes: both the unfiltered Enter and a typed filter select resumed their dormant target (`op-other`, `op-search`) through the same `pendingOperationFocus` route.

fleet-console suite green (537 passed / 8 skipped); removing the deferral turns the new assertion red.

sbluemin:
@codex Please review this PR.

sbluemin:
Both third-pass findings confirmed and fixed in 1ddf7441.

**FIX — Avoid auto-resuming from degraded runtime data.** Correct: only `"pending"` was special-cased, so an open during a stream outage evaluated the retained map or the payload fallback. That also contradicted the axis doctrine already in this codebase — `pluginRuntimeState` deliberately refuses to hand a degraded value to a plugin as current fact. The two unobserved states now part ways on their own terms: `pending` holds the gesture because authority is about to arrive; `degraded` does not, because recovery is open-ended and holding a process launch across it would be worse than not acting. Either way the frame keeps its Resume button.

Verified in the real app by failing `/api/v1/agent/events` outright (6 reconnect attempts, degraded banner up): opening the stowed panel at t=3003ms issued no resume request and left the panel on its Resume card.

**FIX — Check the deferred operation's theater before resuming.** Correct: minimized state is a per-theater axis and the guard read `getCanvasSnapshot()`, so a theater switch during the wait would have asked another theater's list. It now resolves the operation and reads `getTheaterCanvasSnapshot(operation.theaterId).minimized`, and drops the intent when the operation is gone. Reasoned and typed, not separately reproduced — the sequence needs a re-stow plus a theater switch inside the pending window.

fleet-console suite green (537 passed / 8 skipped); reverting the degraded guard turns the new assertion red (5 calls instead of 4).

This is the third review pass. I am ending the loop here under the workflow's stop rule — I have no named reproduced defect that would justify a fourth — and moving to the final audit.

PR #1113, opened 2026-09-16
feat(console): restore Chrome cookie import and fix browser screenshot paste into CLI Operations

sbluemin:
## Summary
- Restore "Import from browser" in the Operation Browser caption. Because the browser is now Desktop-only, the Desktop shell lists the Google Chrome profiles on the machine holding the window, reads the chosen profile's cookies through a headless Chrome over a CDP pipe, and sets them on that Operation's session partition only. Console asks the shell through new `Fleet.*` shell commands on the native-view protocol (`DESKTOP_BROWSER_SHELL_VIEW`) instead of locating Chrome itself. Dialog, glyphs, copy and CSS are the same as before removal.
- Fix "Attach screenshot" / "Attach annotations" into terminal (CLI) Operations: the paste route now receives the PNG, writes it to the OS clipboard of the machine running the CLI (osascript / PowerShell / xclip or wl-copy), and only then presses Ctrl+V. Previously the renderer's `navigator.clipboard.write` landed on the pasteboard after the Ctrl+V reached the PTY, so Claude Code read an empty clipboard and reported no image (reproduced in the dev Desktop; the direct API call with a pre-filled clipboard worked, the button did not).
- Amend the pending `browser-desktop-only` release note: the cookie import is not gone.

Changelog-Amend: browser-desktop-only.md

## Test Plan
- [x] `tsc --noEmit` for fleet-console and fleet-desktop; `pnpm --filter @dotobokuri/fleet-console build`; `pnpm --filter @dotobokuri/fleet-desktop build`
- [x] `vitest run tests/browser-desktop-engine.test.ts tests/server.test.ts tests/instrument-design-contract.test.ts` (console) and the full fleet-desktop suite
- [x] `node scripts/compile-changelog-fragments.mjs --check`, `check:localized-attributes`, `check:core-boundary`
- [x] desktop-e2e (dev Electron 43.1.0, macOS, isolated data dirs, CDP on loopback): before the fix the panel click left the CLI without an image and only 4 s later showed "Image in clipboard"; after the fix the click yields `[Image #1]` in the Claude Code prompt with the terminal toast. Import dialog lists the local Chrome profile, import logged 814/814 cookies into the Operation's partition, and google.com in that Operation's view carries SID/APISID/SAPISID with the signed-in UI.
- [ ] Windows / Linux clipboard writers (PowerShell, xclip/wl-copy) — not verified on this macOS host


sbluemin:

## Addressed
- P2 stale shell-command results — `runtime/fleet-desktop/src/browser-views.ts` — the shell-command handler captures the `session` token when it starts and drops its result/error if `start()`/`stop()` advanced the session meanwhile, so an import that finishes after a Console handoff never reaches the new Console's outbox.
- P2 paste size cap — `runtime/fleet-console/core/host/server.ts` — `BROWSER_PASTE_MAX_BYTES` is now 48 MiB. Measured worst case: a random-noise 3840×2400 PNG is 27.7 MB, 35.2 MiB as base64, which the old 32 MiB cap rejected.

## Validation
- `tsc --noEmit` (fleet-console, fleet-desktop) — clean
- `pnpm --filter @dotobokuri/fleet-console build`, `pnpm --filter @dotobokuri/fleet-desktop build` — success
- fleet-desktop vitest suite — 24 files / 122 tests passed; console `browser-desktop-engine` + `server` tests — passed

PR #1115, opened 2026-09-16
fix(console): keep the Operation Browser off the window's connection budget

sbluemin:
## Summary
- A Desktop window showing a **remote** Console froze on every click: new tab, close tab and navigate in the Operation Browser did nothing, and so did the rest of the Console. The shell's own plumbing was sharing the window's HTTP/1.1 socket pool (6 per origin), so once it filled, every request the page made sat in the queue until a stream closed. Agent-driven browsing kept working because it never leaves the server.
- The Desktop main process now talks to a remote Console through its own in-memory `fleet-shell` session instead of the window's, with the certificate pins installed on both and the session cookie copied from the window's jar (the shell must present the same session, because Console picks the view host by the control holder's handle).
- The Operation Browser panel no longer opens a stream of its own: state changes ride the already-open Operation stream as `browser:state`, and the panel reads its starting point once from `GET /api/v1/browser/operations/:id/state`. The `/stream` route is gone from the API catalog.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console typecheck` / `test` (539 passed, 8 skipped)
- [x] `pnpm --filter @dotobokuri/fleet-desktop typecheck` / `test` (122 passed)
- [x] Console and Desktop builds; `check:core-boundary`, `check:localized-attributes`
- [x] Real remote path (dev Desktop -> LAN TLS Console, paired): `GET .../state` 200 in 3 ms, `POST .../tabs {create}` 200 in 104 ms with the native view created, `browser:state` frames arriving on the shared Operation stream, and the page's spare connections back to 4 (same as a plain browser on a local Console; measured 1 before)
- [ ] Panel repaint observed in the remote window (data path verified there; the React render was not, because the API-created Operation stayed dormant)


sbluemin:
Both findings reproduced against the new fetch-plus-stream seam, and both are fixed in `0c0ea4b28`.

**FIX — refetch after the shared stream reconnects.** Correct, and it was a regression against the old dedicated stream, which pushed current state on every (re)subscribe. The fix is the server side you suggested: a desktop client now receives a `browser:state` frame for every live Operation browser the moment its Operation stream opens. Reconnects run through that same path, so an agent-opened tab or a URL change during an outage no longer sticks. The panel keeps its one-time read because it can mount long after the stream opened.

**FIX — initial snapshot overwriting newer events.** Also correct. The panel now marks that a stream frame arrived and drops the starting-point response in that case, so a slow read cannot roll state back to a value nothing will repeat.

Checks: `@dotobokuri/fleet-console` typecheck and test (539 passed, 8 skipped), Console build.

sbluemin:
@codex review

PR #1119, opened 2026-09-17
fix(console): keep Console shortcuts alive over the Operation Browser view

sbluemin:
## Summary
- The Operation Browser draws its tabs as a native Chromium view inside the Console window, and keyboard focus belongs to one webContents at a time. Once that view held focus — one click on the page was enough — every Console shortcut (Cmd+K, Cmd+P, Quick Launch, the Alt chords) went dead until the user clicked back into the Console.
- Console now declares its chord registry in the browser snapshot; the Desktop shell intercepts exactly those chords on the view, keeps them from the page, and relays them back, and the panel re-presses each one so the existing global handler picks its command. Shortcut policy stays with Console.
- The shell also preserves the keyboard holder across relayed `Input.*` commands. Synthesized input did not move focus on macOS when measured, so this one is a guard against Electron's per-platform focus handling, not an observed repair.
- Protocol changes are additive optional fields (`snapshot.chords`, `relay.keys`), so an older shell or an older Console degrades to today's behavior.

## Test Plan
- [x] `pnpm typecheck` — all workspaces
- [x] `pnpm build` and `pnpm desktop:build`
- [x] fleet-console suite: 125 files, 540 passed / 8 skipped
- [x] fleet-desktop suite: 24 files, 122 passed
- [x] Real app (isolated dev Desktop, macOS arm64, Electron 43.1.0): the mounted panel published the real 14-chord registry into the Desktop snapshot; a Cmd+K delivered to the native view opened the Console operation-search overlay
- [x] Inverse in the same run: Cmd+K never reached the page and was relayed; Cmd+F reached the page and was not relayed
- [x] Measured focus across synthetic input and new-view creation: the Console renderer kept focus in both
- [ ] Windows — not verified; `Mod` maps to Ctrl there


sbluemin:
All four findings **FIXED** in `e509ad64`, each reproduced first and then re-verified in a real isolated Desktop (macOS arm64, Electron 43.1.0).

**P1 — editable-page input for context-gated chords.** Confirmed: `blocksOperationsShortcutWhileEditing` declines non-Alt operation shortcuts while an editor has focus, so `Shift+Digit1` would have swallowed `!` in page forms. The same reasoning applies to a chord the review did not name — `console.undo-close` (`Mod+KeyZ`) is activeElement-gated in `global-shortcuts.ts`, so intercepting it would have replaced page undo with the Console's undo-close. Both are now excluded: only chords carrying Mod, Ctrl or Alt are declared. Verified in the app — the published list no longer contains `Shift+Digit1` or `Mod+KeyZ`.

**P1 — route relayed chords only to the owning Desktop.** Fixed; the fan-out now matches each desktop subscriber's owner against `desktopEngine.currentHost` and skips the rest. State stays broadcast, since state is idempotent and a chord is a command.

**P2 — Ctrl bindings off macOS.** Confirmed and fixed with `foldDesktopChord`: the declared set is folded `Ctrl` → `Mod` on non-Apple platforms before lookup, matching `chordsEquivalent`. This was the one finding that affects the reporting user's platform directly.

**P2 — companion shortcuts.** Fixed; the panel now also declares the active companion chords from `useActiveCompanionShortcuts`. Verified in the app: `Alt+KeyB` and `Alt+KeyA` are published, and pressing Alt+B while the native view holds focus closes the browser companion instead of reaching the page.

Re-verified after the change: Cmd+K delivered to the native view still opens the Console operation-search overlay; `pnpm typecheck`, the fleet-console suite (540 passed / 8 skipped) and the fleet-desktop suite (122 passed) are green.

sbluemin:
@codex review

sbluemin:
Second pass — all four **FIXED** in `942d7c0e`, each reproduced against the code first and re-verified in a real isolated Desktop.

**P1 — AltGr character input.** Confirmed: `matchesChord` splits a `Mod+Alt` chord from a typed character on `event.key`, and a chord rebuilt from the physical code defeats exactly that guard. The shell now applies the same key-versus-code check in `typedCharacterSurvives` before `preventDefault()`, so a layout that reports AltGr as Ctrl+Alt keeps its character and the chord is never relayed.

**P2 — remapped non-Alt operation chords.** Correct, and the modifier filter alone was only incidentally safe because the defaults happen to be Alt or Shift. Operation-group and companion chords are now declared only when the resolved chord carries Alt, so a remap to `Mod+KeyA` stays with the page input.

**P2 — auto-repeat.** Confirmed against the repeat guards in `global-shortcuts.ts:86,95` and `pages/operations.tsx:275,298`. `isAutoRepeat` now rides the relay onto the synthetic event. Verified in the app: `Mod+Alt+KeyZ` delivered with `autoRepeat: true` leaves Zen untouched, and the next fresh press toggles it.

**P2 — relay retry duplication.** Correct; the retried body had no identity for keys. Each intercepted key now carries a monotonic id and the engine skips ids it has already fired, bounded the same way as `executed`.

Also re-verified after the change: Cmd+K into the native view still opens the operation-search overlay, the published registry still excludes `Shift+Digit1`/`Mod+KeyZ` and includes the `Alt+*` companions. `pnpm typecheck`, fleet-console (540 passed / 8 skipped) and fleet-desktop (122 passed) are green.

sbluemin:
@codex review

sbluemin:
Third pass — both **FIXED** in `221f3616`.

**P1 — intercepted-key ids outliving the shell session.** Correct, and this was a regression my own second-pass dedup introduced: the shell resets `keySerial` on every `stop()` while the engine's set persisted across host changes and closure, so a returning shell's first presses looked like retries and were swallowed. The set is now cleared when the hosting shell says `hello` (which is emitted by `start()`, exactly when its counter restarts) and when the engine closes.

Verified against the running server through the real relay endpoint: pressing id 1 emits one `browser:chord`; replaying the identical body emits nothing more (the retry guard still holds); sending `hello` and then replaying id 1 emits again (the session scoping). Before this change the third step stayed silent, which is the reported failure.

**P2 — restoring focus after the user moved it.** Correct; the guard now restores only when the relayed command actually left focus in that view, so a click elsewhere during the in-flight command wins.

Re-verified after the change: a single Cmd+K into the native view opens the operation-search overlay from a closed state. `pnpm typecheck`, fleet-console (540 passed / 8 skipped) and fleet-desktop (122 passed) are green.

This is the third pass and the two findings were both real, one of them a regression from a prior pass, so I am asking for one more look at the final state before merging rather than stopping here.

sbluemin:
@codex review

sbluemin:
Fourth pass — both **FIXED** in `8c32ae55`.

**P2 — matching keyUp.** Correct; only the press was prevented, so a site binding shortcuts to `keyup` could still run them and the page got a release for a press it never saw. Each view now remembers the codes whose press it intercepted and prevents the matching release without relaying it — the command belongs to the press.

Verified in the app with listeners installed in the page: Cmd+K produced **no** `keydown` and **no** `keyup` in the page while the Console overlay opened; Cmd+F produced both in the page and no relay.

**P2 — chord declarations from a non-host Desktop.** Correct; there is one declared set and companion chords follow each window's active Operation, so a non-host window could overwrite the host's. The `place` route now accepts `chords` only when the caller is `desktopEngine.currentHost`; placement itself is unchanged.

`pnpm typecheck`, fleet-console (540 passed / 8 skipped) and fleet-desktop (122 passed) are green.

I am ending the review loop here. Four passes ran, every finding was reproduced against the code and fixed, and the last two passes produced no P1s — the remaining suggestions have been narrowing edge cases rather than defects in the supported path. Further passes would be unbounded.

PR #1123, opened 2026-09-19
feat(console): let Console use put an Operation to sleep and reflect closes on every screen

sbluemin:
## Summary
- Add `console_panel` action `sleep`: an agent can put an idle terminal Operation dormant through the same path as idle dormancy, so it lands on the Ended shelf and resumes with its captured session. Refused for the caller itself, chat Operations, running/awaiting/background targets, and Operations without a captured provider session.
- Broadcast `operation:removed` when an Operation enters deferred deletion and `operation:changed` on restore; the client drops/re-adds the row from those frames, so an agent's `close` (and any other window's) shows up immediately instead of waiting for the next re-hydration.
- Cover the sleep guards in the Console use surface test; release note added.

## Test Plan
- [x] fleet-console vitest suite (540 pass)
- [x] Isolated Console e2e with a `codex--gpt-5.6-luna` caller: `sleep` → target `status: dormant`, panel shows Ended/Resume; `close` → `operation:removed` frame and the row leaves the sidebar without reload (0 page errors); restore via API → `operation:changed` brings it back.


sbluemin:
## Review dispositions (pass 1)

**DEFER — `server.ts` restore branch broadcasts only `operation:changed`, so a window that connected after a Theater was forgotten gets orphaned Operations on Theater undo.**

The gap is real but predates and sits outside this PR: Theater lifecycle (`theater:forgotten` / `theater:restored`) is published only in-process and has never reached the browser stream, so a second window never learns a Theater was forgotten or restored either way. This PR's scope is Operation removal/restore reflection. Its effect on the Theater-undo path is not a regression: the sidebar scopes rows by `activeTheaterId` (`operations-side-bar.tsx:413`), so an Operation whose Theater is unknown to that window is not rendered — the same visible outcome as before (absent until refresh) — and a window that still holds the Theater now correctly gets its Operations back, which it previously did not. Broadcasting Theater lifecycle (Theater + groups + Operations, or an authoritative rehydrate signal) is a separate change and will be tracked as a follow-up rather than absorbed here.

No FIX-class findings in this pass; proceeding to final audit.

PR #1127, opened 2026-09-19
feat(console): choose a temporary session or a persistent profile in the Operation Browser

sbluemin:
## Summary

- The Operation Browser caption gains a session marker: **임시 (temporary)** or **내 프로필 (my profile)**. Temporary stays the default — agents drive this browser, so signed-in state is asked for, not inherited. Switching closes the open tabs (a session cannot be swapped into a live view), and the marker's menu can clear the profile.
- One global profile is shared by every Theater and Operation. It opens through `session.fromPath()` instead of a `persist:` partition, which keeps it out of Electron's `userData` — that path is `%APPDATA%` (Roaming) on Windows, where a browser cache and cookie jar would sync on every logon. Profiles live under the Fleet root (`<FLEET_DESKTOP_DATA_DIR ?? ~/.fleet/desktop>/browser-profiles/<id>`), so the app's own session needs no migration, isolated dev runs are isolated for free, and no platform branch is needed. The Console sends a profile id and the shell builds the path; no path crosses that boundary.
- Chrome cookie import targets whichever session the Operation uses and the dialog names the destination. Import stays unavailable on Windows (unchanged). Clearing a profile closes its views first, because Windows cannot delete open files.
- `DesktopBrowserView` gains `profile` beside the existing `partition`, so an older Desktop shell that does not read it falls back to a temporary session rather than failing.
- Also fixes the Chrome profile select overflowing a narrow import dialog by 74px — the SDK control carries a `min-width: 220px` floor that predates this change (#1113).

## Test Plan

- [x] `pnpm typecheck`, `pnpm build`, `pnpm --filter @dotobokuri/fleet-desktop build`
- [x] `pnpm test` — Console 125 files / 540 tests, Desktop 24 files / 122 tests, all passing
- [x] Real app (isolated dev Desktop, native Chromium views): set a cookie under **내 프로필** → `COOKIE:fleetprofile=kept-1`; switch to **임시** → `COOKIE:none` (isolated); switch back → cookie intact (persisted); clear profile → `COOKIE:none`
- [x] Profile directory created at `<isolated>/desktop/browser-profiles/default` (1.8 MB, `Cookies` + `Local Storage`) — outside Electron `userData`
- [x] Switch confirmation dialog names the tab count; cancel restores; import dialog shows the destination row; select no longer overflows (−17px, inside)
- [ ] **Unverified — requires Windows**: Roaming avoidance, file locking on clear, MAX_PATH
- [ ] **Unverified — requires Linux**


sbluemin:
## Review pass 1 — dispositions

All three findings reproduced against the supported path before changing anything. All three **FIX**, pushed in `3adc2bcfd`.

**P1 · Dispose temporary contexts when the last tab closes — FIX**
Reproduced: `closeTab()` kept `op.contextId`, so a tab reopened inside the 5-minute idle window came back to the same Electron partition. The marker's own copy promises the opposite ("탭을 닫으면 로그인이 사라집니다"), so this was a gap this PR created. The context is now disposed when its last tab closes. A persistent profile is disposed on the same path and loses nothing — its cookies live in the profile on disk, not in the context.
Verified in the real app: temporary tab sets a cookie → close last tab → reopen immediately → `COOKIE:none`. Inverse under the profile → `COOKIE:fleetprofile=kept-1`, unchanged.

**P1 · Accept browser snapshots from older Consoles — FIX**
Correct, and it is the compatibility direction opposite to the one I had handled. The shell's `spec.profile ?? null` fallback was unreachable because the predicate rejected the whole snapshot first. An absent field now reads as `null`, and `profile` is optional on the type.

**P2 · Schedule engine shutdown after profile resets — FIX**
Confirmed: `setProfile()`/`clearProfile()` close tabs without passing through `closeTab()`, so neither armed the idle timer and the engine could hold the Desktop connection open indefinitely. Both now call `scheduleIdle()`.

Checks after the fixes: `pnpm typecheck`, `pnpm build`, Desktop build, and `pnpm test` (Console 125 files / Desktop 24 files) all pass.

Still unverified and unchanged from the PR body: Windows real hardware (Roaming avoidance, file locking on clear, MAX_PATH) and Linux.

@codex Please review this PR.

sbluemin:
## Review pass 2 — dispositions

Both findings **FIX**, in `e552b3e8b`, together with a UI revision the author asked for.

**P2 · Hide the native view while the profile menu is open — FIX**
Correct, and the panel's own module comment already states the invariant this broke: "네이티브 뷰는 언제나 페이지 위에 그려지므로, 위에 떠야 하는 것은 뷰를 물러서게 해야 보인다." Measured with a tab open: the menu extended 164px past the native view's top edge, so 내 프로필 and 내 프로필 비우기 sat behind the page.
Not fixed by adding `aria-modal` — a menu is not a modal and saying so would lie to assistive tech. The caption now publishes its open menus through the panel store (the bridge the tab strip already uses) and the panel treats that as covering, immediately rather than waiting out the 200ms placement poll. The viewport menu was covered the same way and is fixed by the same signal.
Verified: opening the menu posts `place {visible:false}`, closing posts `{visible:true}`.

**P2 · Serialize last-tab disposal with tab creation — FIX**
Confirmed by reading `createTab()`: the tab is inserted at the end, after `Target.createTarget` has awaited view attach (up to 10s), so a concurrent close of the last tab saw zero tabs and disposed the shared context. In-flight creations are now counted and the context is kept until none remain.

### UI revision (author-directed, not review-induced)
The session marker is now a glyph alone — no label, no pill, no border — so it sits in the caption like the viewport control rather than reading as a button pinned onto the panel. Cookie import moved into that menu: cookies land in whichever session the marker names, so the action belongs with the session control, and the caption loses a button. Each menu row leads with its own glyph and its description is one short line.

Checks: `pnpm typecheck`, `pnpm build`, and `pnpm test` (Console 125 files / Desktop 24 files) pass. Windows and Linux remain unverified as stated in the PR body.

This was the second pass and no finding remains open, so I am closing the review loop here and moving to the final audit.

PR #1130, opened 2026-09-19
fix(console): hand oversized browser screenshots to agents as files

sbluemin:
## Summary

- The Operation Browser returned every screenshot inline as a PNG. A calling Agent CLI caps MCP tool results by tokens and counts base64 image data against that cap, so a full-viewport capture was spilled to a file by the CLI and the model never saw the image. Captures are now JPEG, and one that still exceeds the inline budget is written to the Console host — where the calling agent runs, even when the Desktop drawing the views is remote — and returned as a path. Resolution is never reduced: measurement showed q80 at full resolution stays legible while a downscale destroys small UI text and Hangul.
- Screenshot files stay owner-readable, bounded per Operation, and are reclaimed when the browser is revoked or the server shuts down.
- `console-e2e` Desktop references: the only working launch line lived in the Shell/CDP lane, so a Desktop run made for any other claim never reached it, and that line skipped the isolated data root and read the user's real `~/.fleet`. The launch moves to Setup, which every Desktop run reads.

## Test Plan

- [x] `pnpm typecheck` (fleet-console)
- [x] `pnpm build` (fleet-console) — change present in `dist/fleet.mjs` and `dist/cli.mjs`
- [x] `npx vitest run` (fleet-console) — 126 files, 542 tests, 0 failures
- [x] New `tests/browser-screenshot-budget.test.ts`: oversized capture returns a path and no image block, saved bytes equal the capture, mode 0600, removed on revoke; a capture within budget stays inline
- [x] Desktop E2E in the real app (isolated dev Desktop from this worktree, Electron 43.1.0): added the Codex GPT-5.6-Luna gateway model, launched an Operation on it, and had that agent call `mcp__fleet-browser__computer`. It received the path, read the file, and correctly reported the page title from the image. The screenshot namespace was removed when the app quit.
- [x] Changelog fragments validated (`compile-changelog-fragments.mjs --check`)

## Known follow-up (not in this PR)

The E2E run showed the capture is not normalized to CSS pixels: a 1280x800 CSS viewport produced a 2560x1600 JPEG on a Retina display, because `viewport.scale` stays 1 at capture time. The extra pixels are useless for CSS-pixel coordinates but quadruple the bytes, so in practice almost every screenshot now takes the file path. Normalizing that capture would let most screenshots stay inline.


sbluemin:

## Addressed
- P2 release-vs-in-flight capture — `runtime/fleet-console/core/host/browser/tools.ts:66` — confirmed the mechanism: `agentCall` checks the abort only after the tool callback returns (`browser/service.ts:401`), and `revokeOperation` does not await `closeOperation`, so a capture finishing after the reclaim recreated the directory and kept the page copy until shutdown. The aborts do run synchronously before `release()`, so the fix is at the write itself: a cut-off call returns the capture inline and writes nothing.
- P2 four-digit serial overflow — `runtime/fleet-console/core/host/browser/screenshot-store.ts:53` — reproduced the ordering: with 13 files from `shot-9989` to `shot-10001`, a lexical sort selects `shot-10000.jpg` for deletion, i.e. the newest capture, after which `save()` returns a path that no longer exists. Pruning now orders by the parsed serial, which selects `shot-9989.jpg`.

## Validation
- `tests/browser-screenshot-budget.test.ts` — new case "writes nothing when the call was already cut off"; verified it fails without the guard (`expected false to be true`) and passes with it.
- `pnpm typecheck` (fleet-console) — pass.
- `npx vitest run` (fleet-console) — 126 files, 543 tests, 0 failures.
- `pnpm build` (fleet-console) — pass.

## Notes
- No declines or deferrals in this pass; both findings were reproduced before editing.
- The Retina capture normalization noted in the PR description stays out of this PR.

sbluemin:

## Addressed
- P2 stale cleanup before lock ownership — `runtime/fleet-console/core/host/browser/screenshot-store.ts` — reproduced the ordering: the store is built in `createConsoleServer()` (`core/host/server.ts:694`) while `lock.writeLock(...)` only runs at the end of `start()`, so a second console for the same data directory wiped a serving console's namespace, and its failed-start `cleanup()` wiped it again. Leftover reclamation now happens on the first `save()`, when this process is already serving, and `cleanup()` removes only the directories this process created, dropping the namespace root only when nothing else remains in it.

## Validation
- `tests/browser-screenshot-budget.test.ts` — new case "leaves a serving console's screenshots alone when a second one starts and gives up"; verified it fails against the previous construction-time wipe (`expected false to be true`) and passes with the fix.
- `pnpm typecheck` (fleet-console) — pass.
- `npx vitest run` (fleet-console) — 126 files, 544 tests, 0 failures.
- `pnpm build` (fleet-console) — pass.

## Notes
- No declines or deferrals in this pass.
- A crashed console's residue is now reclaimed by the next run's first spill rather than at its construction; the files stay owner-only in the OS temp directory until then.

PR #1132, opened 2026-09-19
feat(desktop): update the shell itself from GitHub Releases

sbluemin:
## Summary

- Fleet Desktop now replaces itself from the GitHub Releases it already ships to, instead of sending the user out to the release page. It checks on a fixed interval, starts the download only when the user asks for it, and installs only when the user restarts — nothing is fetched in the background and nothing installs on quit.
- Updater metadata is part of the release again: `electron-builder` keeps `latest*.yml`, the release verifier requires it rather than forbidding it, and a Desktop-unchanged release carries it forward so its update check does not 404. Artifact names lost the space that desynchronized GitHub's asset naming (`Fleet.Console-…`) from the updater's download URL (`Fleet-Console-…`).
- Updates are announced in a bubble on the help button. When the shell and a local managed Console are both behind, one restart installs both and the Console row folds into it; a remote Console is a different machine, so it never folds. The Desktop row is split — the name opens the release page, the chip performs the update.

## Test Plan

- [x] `pnpm typecheck` (workspace), `pnpm check:core-boundary`
- [x] `pnpm --filter @dotobokuri/fleet-console test` (540 passed / 8 skipped), including `instrument-design-contract`
- [x] `pnpm --filter @dotobokuri/fleet-desktop test` (120 passed)
- [x] `node --test scripts/*.test.mjs` (64 passed)
- [x] Console notice + help-menu rows verified in an isolated Console in a real browser
- [x] Desktop shell verified in an isolated dev app against a local update feed: boot → check → download (progress) → install → relaunch
- [ ] **Unverified:** relaunch of a signed/notarized packaged build after install; Windows unsigned NSIS update on a real Windows host; Console self-update regression on a published channel

## Notes

- The Windows job no longer signs through SignPath. Nothing may rewrite the installer after packaging: the hash recorded in `latest.yml` belongs to the file that build produced, and a substituted file fails the updater's integrity check. The repository never held `SIGNPATH_*` values, so that path had never run.
- `electron-updater` is bundled into the shell entry, so the shell-only ASAR contract is unchanged. The ESM output needed a real `require` at the entry — the bundle builds without it but throws on the first frame.


sbluemin:
@codex Please review this PR.

PR #1134, opened 2026-09-19
fix(console): hand every browser screenshot to agents as a file

sbluemin:
## Summary

- Operation Browser screenshots no longer ride in the tool result as base64. Base64 is priced as text, so one 1386x1163 capture cost **~35,000 tokens** — measured as the context delta across the call — and the model never saw an image, only the base64 string. Every capture is now written to the Console host and the result carries only its path; the agent reads the file to get a real image block when it needs one.
- The 64,000-char inline budget is gone with it. That threshold had the cost relationship backwards: a capture *over* budget went to a file for ~450 tokens, while one *under* budget was inlined for ~35,000.
- Screenshot-bearing returns now declare `isError`. Without it `toMcpCallToolResult` fails its shape check and `JSON.stringify`s the whole structured result into one text block — which is what turned the image block into text in the first place. `browser_batch` had the same gap.

### Measured

Same metric on both sides: per-turn `input + cache_read + cache_creation` delta.

| | before | after |
|---|---|---|
| tool result, per capture | 35,064 tok (inline) | **495 tok** |
| agent actually sees the screen | 35,064 tok — and it does not | **3,927 tok** — and it does |

In the session that prompted this, 5 inline captures took **175,318 tokens**: 94.6% of all `computer`-call cost and 36% of that session's 493,350-token context.

## Test Plan

- [x] `tsc --noEmit` on fleet-console
- [x] `browser-screenshot-handoff.test.ts` 3/3 — replaces `browser-screenshot-budget.test.ts`; asserts the result carries no image block and no base64, the file is byte-identical and 0600, and the browser reclaims it
- [x] `browser-desktop-engine.test.ts` 1/1
- [x] fleet-console build
- [x] Real app: isolated Electron Desktop from this worktree, Claude Code v2.1.278 / Sonnet 5 Medium Operation drove the Operation Browser — 6 `computer` calls, 6 file reads, task completed correctly (had to read a number off the screen to answer), no inline base64 in any result


sbluemin:
@codex Please review this PR.

PR #1141, opened 2026-09-19
fix(gateway): stop Cursor turns from discarding a generation on a native read

sbluemin:
## Summary
- Name the advertised read tool in the always-applied Cursor rule. The model opened with a Cursor-native file read in 9 of 10 measured baseline trials even with the caller's `Read` advertised; that read fail-closes, so each one cost a whole discarded generation. Native read attempts went to zero, and 5-trial mean wall time dropped 15218ms → 13771ms on Grok 4.6 and 21427ms → 14736ms on Opus 5 with caller requests, tool calls, errors, and task outcome unchanged.
- Keep tool arguments on a non-streaming turn. The Cursor adapter carries the whole argument JSON on the closing item and emits no `function_call_arguments` events; `encodeAnthropicSse` already accepted that shape, so `collectAnthropicMessage` now does too instead of handing the caller an empty tool input.
- Accept a `system` sent as a plain string. Anthropic allows it, but every reader in the package assumed the block array and reached `.filter`/`.map` on a string, failing the request on every provider. The Anthropic wire reduces it to that array once, and refuses an unusable shape as a 400.

Also measured and rejected: enabling the already-implemented `read-direct` redirect for the same native read. It answers with a body the native shape cannot certify as complete, and the model re-read instead — Grok 4.6 went 15218ms → 20116ms and `Read` calls 5 → 15. The dispatch-site exclusion is intentional and stayed.

## Test Plan
- [x] `pnpm --filter @fleet-console/ai-gateway test` (198 passed)
- [x] `pnpm --filter @fleet-console/ai-gateway typecheck`, `pnpm typecheck` (repo-wide)
- [x] `pnpm --filter @fleet-console/ai-gateway build`
- [x] Both new tests confirmed failing with the source fixes stashed and passing with them
- [x] Live `e2e:provider-loop` against `cursor--grok-4.6` and `cursor--claude-opus-5` at effort low: 2/2 successful trials each, 3 caller tool calls, 0 caller errors, router and loopback cleanup verified. This runner could not complete before the non-streaming argument fix.
- [x] Live read/search/edit workload, 5 trials per model before and after, frozen model/effort/prompt/catalog/theater: native `readArgs` frames 4/5 and 5/5 trials → 0, task outcome 5/5, 0 caller errors, bridge reuse `exact_match` on every continuation
- [x] Rule text checked offline for both sides: a catalog with no read tool, an unrelated-tool catalog, and an empty catalog are byte-identical to before
- [x] `node scripts/compile-changelog-fragments.mjs --check`, `node scripts/changelog-korean-seed.test.mjs`, `node scripts/check-core-boundary.mjs`, `git diff --check`

Not verified: no real Console Operation or Desktop run — the change is contained to the `@fleet-console/ai-gateway` package.

Made with [Cursor](https://cursor.com)


sbluemin:
@codex Please review this PR.

PR #1142, opened 2026-09-20
fix(browser): keep native view visible beside context menus

KIMMUSIC:
## Summary
- Ignore explicitly marked transparent context-menu backdrops when measuring native Browser occlusion, while retaining actual menu-card and modal occlusion.
- Dismiss operation/group/Theater menus on renderer blur, and restore trigger focus only while the Console document still owns focus.
- Keep this PR open for manual merge. **Do not enable auto-merge.**

## Test Plan
- [x] Console and Desktop builds; final renderer rebuild and Console typecheck.
- [x] Existing browser Desktop engine test and `git diff --check`.
- [x] Windows x64 Electron 43.1.0, isolated worktree build: operation right-click keeps Browser visible; removing/restoring the backdrop marker reproduces `visible: false` / `true`.
- [x] Actual menu-card overlap still occludes; Escape restores visibility; real search modal hides/restores Browser.
- [x] Keyboard menu opening and Escape focus return; native CDP target focus/click does not steal focus back.
- [x] User tested the development Desktop and confirmed the change.
- [x] Final independent code review: no actionable findings.
- [ ] Existing screenshot-handoff suite is not fully green on Windows: file permission assertion expects `0600`, receives `0666` (`browser-screenshot-handoff.test.ts:57`). Combined targeted run: 3 passed, 1 failed. No permission tests were changed.

Automated native-pointer verification used CDP target activation, not an OS-level physical click. The user subsequently performed manual verification; no more detailed manual assertion breakdown was recorded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sbluemin:
Thanks for the careful fix, @KIMMUSIC.

Marking the transparent backdrop instead of the whole overlay subtree is the right cut: the click catcher stops falsely occluding the native view while the menu card itself still does. The `blur` dismissal and the `document.hasFocus()` guard also close the gap where a native-view click never reaches the renderer's pointer handler.

The `0600`/`0666` screenshot-handoff assertion is a pre-existing Windows permission-model difference and unrelated to this change. Merging now.

PR #1145, opened 2026-09-20
feat(console): pick built-in subagents from a grouped checklist

sbluemin:
## Summary
- Replace the seven bare switches in the Claude Code harness card with a grouped checklist: each built-in subagent carries a one-line description, group headers hold the mixed state, and the approval gate is now the card's only switch. A switch answers "turn this one setting on or off"; this row asks "which of these may a session use".
- Fleet owns the descriptions because the CLI reports names only and exposes no surface that describes them. Names Fleet does not know still render, under their own group, so a CLI update never hides something the user can turn off. Wording comes from each agent's `whenToUse` in the installed bundle (measured on 2.1.278).
- Settings writes now coalesce per field instead of refusing while one is in flight, disabled names stay listed and reversible when the roster cannot be read, and the shared checkbox glyph draws a check mark rather than an upward chevron.

## Test Plan
- [x] `pnpm typecheck`, `pnpm build`
- [x] `pnpm test` — Console 587 passed / 8 skipped, Desktop 120 passed
- [x] `pnpm check:core-boundary`, `pnpm check:localized-attributes`
- [x] Replaced the superseded save-lifecycle case: in-flight writes coalesce, the intermediate intent is never sent, and a failed field rolls back alone
- [x] Real app (isolated Console, agent-browser fallback — the Fleet Browser pane reported a 0×0 viewport): default / mixed / degraded states captured; turning `Plan` off flips the group to mixed and persists; three clicks during an in-flight save all land; group header toggles both directions; with `claude` off PATH the disabled name stays listed and can be re-enabled
- [ ] Not covered: cleaning up names a CLI update removed, and progress detail for "Read again" beyond its pending label


sbluemin:
@codex Please review this PR.

PR #1149, opened 2026-09-20
feat(console): let chat Operations rest and resume like terminal ones

sbluemin:
## Summary
- Chat gains the dormancy the terminal surface already had: the SDK child and journal are collected, the activity axis releases its owner, and the surface marker stays so resume returns to the conversation rather than a terminal. Restored chat Operations stay dormant after a console restart until resume or the first message wakes them, and the idle sweeper now reaches the chat adapter (skipping a running turn, a pending question, or live background work).
- Ctrl+C in a chat panel arms a hint and a second press parks the Operation. The gesture follows the active Operation rather than input focus, a standing selection keeps the copy shortcut, key repeat never confirms, and the next keystroke or leaving the active seat disarms it.
- `console_panel` sleep and resume accept the chat surface; the "a chat Operation is never dormant" doctrine strings are gone.

## Test Plan
- [x] `pnpm --filter @dotobokuri/fleet-console typecheck && build && test` (590 passed, 8 skipped)
- [x] `pnpm --filter @fleet-plugins/scuttlebutt typecheck && test`
- [x] Real browser, isolated Console: Ctrl+C twice parks the panel (hint first), resume replays the conversation and the resumed model still recalls the earlier turn
- [x] Real browser: gesture fires with focus on `body` while the panel is merely active; with two chat panels only the active one parks; a standing text selection copies instead
- [x] Real browser: console restart restores the chat Operation dormant (canary host build restored it live, for contrast)
- [x] Real browser: idle sweep at a 1-minute threshold parks an open chat panel (~85s) and the live panel flips to the dormant card


sbluemin:
@codex Please review this PR.

PR #1150, opened 2026-09-20
feat(console): ask where to open a link from an Operation's CLI or chat

sbluemin:
## Summary
- Clicking an http(s) link in an Operation's CLI or chat view opens a chooser card at the pointer: **Fleet Browser** opens that Operation's browser companion on the page (reusing a blank tab, otherwise a new one), **My browser** hands it to the default browser. The card also replaces the terminal's native confirm; the global Shell, which has no card, keeps it.
- Addresses the CLI prints as plain text are now clickable at all (the web-links addon), because Claude Code emits no OSC 8 for this terminal. OSC 8 links and plain URLs pass through one gate that admits http(s) only.
- Where the Fleet Browser cannot open — a Console in a browser tab or on a phone, or while another screen has paused it — nothing is asked and the link opens exactly as before.

## Test Plan
- [x] `npx vitest run` in `runtime/fleet-console` (137 files, 590 tests)
- [x] `node scripts/compile-changelog-fragments.mjs --check`, `check-localized-attributes`, `check-core-boundary`
- [x] Electron Desktop (isolated dev instance): CLI link hover underline → click → card → Fleet Browser opens the panel and loads the page; chat markdown link → card → new tab beside the existing one; My browser → OS default browser, no Fleet tab
- [x] Keyboard: default row focused but unpainted, ↑↓ moves and reveals the ring, Enter opens, Esc closes; hover treatment identical on both rows
- [x] Paused engine (second non-Desktop client attached): no card, link opens as before
- [x] Console opened in a browser tab: chat link → plain new tab, CLI link → existing confirm → new tab


sbluemin:
@codex Please review this PR.

PR #1151, opened 2026-09-20
refactor(console): fold chat background work into the lines already there

sbluemin:
## Summary

The background-work surfaces each took a row of their own, so one job cost the conversation a permanent 33px and seventeen jobs measured 1,097px in a sheet that is 62% of the panel.

- **Ledge → composer meta row.** The meta row becomes a three-column grid (coordinate · ledge · width glyph); its middle cell was always empty. The ledge's own 32px row and hairline are gone. While a notice borrows the coordinate slot the ledge drops the job title but keeps the count and the door.
- **Jobs fold into the tally.** A job-spawning call is no longer an anchor in the ledger body; it folds into a `job` family clause that names its own running and failed counts, so the exceptions are never hidden by the fold. The door to a specific job lives in the tally's expansion.
- **Tally expansion is one time-ordered tree.** Leaves drop the per-step box (33px → 24px) and hang off a single hairline stem. Families are not branched — that would lose "what followed what" — so each leaf carries its own family glyph. Standalone steps in the ledger body keep their box; that box is what says "this did not fold".
- **Work sheet: rows under kind branches.** Cards (57px, two lines) become 28px rows; the card's second line moves to the row tail. Settled jobs group into collapsible subagent/shell/workflow branches, each head keeping its failure count while closed. A single-kind session gets no branch heads.
- **Fixed:** the sheet's settled section counted `!open` — failed, stopped and unknown jobs included — while labelled "Done". It now says "Ended".

`instrument-design-contract` is updated together with the grammar it pins.

## Test Plan

- [x] `vitest run` — 137 files, 590 tests pass (design contract updated to the new grammar)
- [x] `tsc --noEmit`, Console build, `check:core-boundary`, `check:localized-attributes`
- [x] `compile-changelog-fragments --check`
- [x] Real app: isolated Console on this build, chat Operation with 5 live background jobs (4 shells incl. 1 failure + 1 subagent) — verified the inline ledge, the `작업 N건 넘김 · 1건 실행 중 · 1건 실패` clause, the expanded tree, and the sheet's kind branches collapsing while keeping their failure count; checked in both Instrument and Whites
- [ ] Not verified: screen-reader announcement order and keyboard-only traversal after the ledge moved into the meta row; Desktop shell


sbluemin:
@codex Please review this PR.

PR #1156, opened 2026-09-20
refactor(console): retire the gateway model guard hook asset

sbluemin:
## Summary
- Retire `fleet-gateway-model-guard.mjs`: the PostToolUse `Workflow` receipt contract is removed, and the SessionStart version stamp now lives in `hooks/hooks.json` as an exec-form command hook that writes the rendered response to stdout, with the payload passed as an argument rather than embedded in evaluated code. Routing itself already moved to the routing Mod.
- Fix the AI Gateway Routing toggle, which never reached the client settings contract: the state projection dropped `delegationRoutingEnabled` (so the switch always drew as off) and `toSettingsUpdate` had no branch for it, so the trailing fallback wrote `agentIdleDormantMinutes` instead. This also repairs the `pnpm typecheck` failures that landed with #1152.

## Test Plan
- [x] `pnpm typecheck` (green; was failing on canary before this branch)
- [x] `pnpm test` — all workspaces pass
- [x] `pnpm build`
- [x] `check:claude-agent-sdk-boundary`, `check:core-boundary`, `check:core-unified-agent-absent`, `check:localized-attributes`
- [x] Rendered plugin snapshot inspected: guard asset gone, no `PostToolUse`, SessionStart hook inline
- [x] Real Claude Code session with the same exec-form hook injected `Fleet plugin version: <version>` into session context
- [x] Isolated Console in a real browser: Routing toggle off → store writes `delegationRoutingEnabled: false` with `agentIdleDormantMinutes` untouched, survives reload, and toggling back on restores the default


sbluemin:
@codex Please review this PR.

PR #1166, opened 2026-09-20
feat(gateway): assign delegated models with Jev or AI decisions

sbluemin:
## Summary
- Add Jev and AI-model routing using delegated task, model constraints, benchmark and quota data; retain local rule-based assignment when disabled, on failure, and for Workflow stages.
- Run AI judges in fresh non-persistent, tool-free SDK sessions without a Fleet system prompt; default to Sonnet at low effort and offer Opus plus enabled Gateway models.
- Add editable routing-test presets and clear results, reuse the shared model picker, and simplify TypeSafe service/auth presentation.

## Test Plan
- [x] Gateway typecheck, build, and all 218 tests
- [x] Console typecheck, build, and all 578 passing tests (8 existing skips)
- [x] Isolated Fleet Browser: routing toggle, method/model controls, TypeSafe rows/signout, presets and sentence-separated help
- [x] Real loopback HTTP cancellation and normal delayed assignment tests
- [x] Changelog validation and git diff checks
- [ ] Additional live-provider quality/latency measurements (not run; user exercised the handoff)

## Product boundaries
Workflow stages remain rule-based. Off means local assignment without AI decision calls. Jev must not prefilter candidates by deterministic tiers/priority/quota. AI-model calls are fresh and non-persistent; no tools, plugins, ambient MCP, or Fleet system prompt. Test requests incur real decision quota but do not execute the selected worker.


sbluemin:
@codex Please review this PR.

sbluemin:
@codex Please re-review the latest head. The user clarified that AI routing must default OFF on installation/update. Absent and legacy false now use local rule-based fallback; only explicit true enables AI calls. Verified fresh OFF, explicit ON, unrelated-write persistence, OFF, and legacy false with the real settings store. Gateway 218 tests, Console settings 9 tests, typecheck and affected builds pass.

sbluemin:
Review disposition: DECLINE for discussion_r4057192572. The production catch at routing-decision.ts:72 checks `options.signal?.aborted` before `isCallerAbort`, so a plain Error("client disconnected") is already rethrown without fallback. Reproduced with the built public decision API, an in-flight model chooser aborting the caller signal with that exact Error, and a shared providerLoad Map: the same Error propagates and the Map remains empty. No code change is warranted; the suggested failure does not reproduce on this head.

PR #1167, opened 2026-09-20
feat(console): let users divide companion panel widths

sbluemin:
## Summary
- Add a divider between Session / Session Analyst / Browser slots in the Cruise companion layout: pointer drag, `←`/`→` 24px steps, double-click to return to equal. Widths are per-panel weights kept in local storage, so a panel opens at the width you gave it; a newly opened panel enters with the average share instead of upending the neighbours.
- Keep a 320px floor per slot, and let the floors retreat proportionally when the arena cannot hold them, instead of freezing the divider. Tactical and War Room keep their own grids and get no divider.
- Fix closing one companion closing the others. Leaving companion layout is now decided once by the host, only when no panel remains visible; the caption chips, `Alt+A`/`Alt+B`, and the Analyst's Escape path just hide their own panel.

## Test Plan
Verified in an isolated Fleet Desktop (Electron 43.1.0, macOS arm64) built from this branch, headed, 1700x1000 window.

- [x] 3 slots start equal (437/437/437); dividers render in the 8px gaps
- [x] Drag exchanges width with the adjacent slot only (437/437/437 -> 555/320/437, Browser untouched) and clamps at the 320px floor
- [x] `←`/`→` on a focused divider moves 24px per press and touches only its pair
- [x] Double-click on a divider restores equal widths and clears the stored widths
- [x] Widths persist across reload; reopening one companion keeps the ratio, and adding the second gives it the average share (555/320/437)
- [x] Narrow window (1100px) retreats the floors proportionally with no overflow or collapsed slot
- [x] No dividers in Tactical or War Room
- [x] Closing Session Analyst keeps Browser open, and the inverse; closing the last panel leaves companion layout
- [x] Same through `Alt+A` and through Escape in the Analyst composer
- [x] Divider resolves theme tokens in instrument, carbon and whites (light)
- [x] `typecheck`, `test` (583 passed), `check:core-boundary`, `check:localized-attributes`, `boundary:guard`, console + desktop `build`, changelog `--check`

No new permanent test: this is layout geometry and panel visibility, which the admission policy keeps out of the suite, and the sibling width/geometry helpers (`pane-geometry`, expanded-surface `fitMinimums`) carry no unit tests either. Behaviour was verified in the real app as recorded above.


sbluemin:
@codex Please review this PR.

sbluemin:
All three findings reproduced and fixed in `001d9d2f4`.

**P1 — Normalize unchanged slots before writing pixel weights — FIX**
Reproduced. Stored weights `555:320:437`, arena widened to 2600 (rendered `1093/630/861`), then a 100px drag on the first divider: the untouched Browser moved **−338px** to 523 and the divider landed **234px** from the pointer. The dragged pair was written in the current arena's pixel scale while the third slot kept the old scale, and normalization mixed them.
Now every slot is rewritten from its current rendered width, which is what the expanded-surface divider already does with its full pane array. Re-verified in the app at 2400px: `851/491/671` → drag +100 → `951/391/671`, Browser unchanged, divider centre exactly on the pointer.

**P2 — Preserve slot floors when enough width is available — FIX**
Reproduced. Two panels made asymmetric (`900:320`), then the third opened: content 1304 ≥ 3 × 320, yet Analyst rendered at **299px**. Raising a short slot to the floor and then scaling everything back down undid the raise.
Now slots below the floor are pinned at 320 and only the remaining width is shared by weight among the rest. Re-verified in the app through the same UI path: `594/320/398`, nothing below 320, sum exact. The proportional retreat when the arena genuinely cannot hold every floor is unchanged (728px arena → `301/174/237`), and the boundary case content == count × 320 lands on `320/320/320`.

**P2 — Disable panel geometry transitions while resizing — FIX**
Confirmed: `.canvas-operation` transitions `left`/`width` at `--duration-slow`, and the resizing attribute only set `user-select`/`cursor`. The repo already cuts this glide for the two analogous gestures (`.canvas-operation.is-dragging`, `body[data-side-bar-animating]`), so the divider followed the pointer while the panels kept re-easing behind it.
Added `body[data-companion-resizing="true"] .canvas-operation { transition: none; }`. Verified mid-drag in the app: `transitionProperty: none`, `transitionDuration: 0s`, and the first pane's right edge equal to the divider's x (859 = 859); after release the attribute is gone and the normal `0.36s` glide is restored.

Checks after the fixes: `typecheck`, `test` (583 passed), console + desktop `build`.

sbluemin:
@codex Please review this PR.

sbluemin:
Second pass: two FIX, one DECLINE. `2fe0de3f1`.

**Keep narrow-layout dividers movable — FIX**
Reproduced, and worse than reported. The clamp floor fell back to half the pair, so the allowed range collapsed to a single point: at a 1100px window with three panels (pair 475px) a 120px drag produced **−64px** — the first input snapped the unequal pair to equal and then nothing moved at all. At arena 640 (two panels) and 950 (three) the movement was exactly 0px.
Now the floor stays 320 whenever the pair can hold two of them, and only falls back to a quarter of the pair when it cannot, leaving the middle half as travel. Verified in the app at 1100px: `301/174/237` → drag +70 → `356/119/237` → drag −70 → `286/189/237`, Browser untouched throughout. The wide case is unchanged: at pair 869 the floor is still 320 and the travel is still ±115.

**Clean up resize state when pointer capture is lost — FIX**
Reproduced: holding the divider and pressing `Alt+T` unmounts it, and after releasing the pointer `data-companion-resizing` stayed on `<body>` with `user-select: none` and `cursor: col-resize` — it survived further clicks and mode changes.
The first attempt, a `lostpointercapture` listener on the divider, did **not** fix it, so I instrumented the page instead of guessing. Chromium dispatches `lostpointercapture` at the **document**, not at the element, and defers it until the next pointer event; the following `pointerup` targets whatever is under the cursor (`canvas-triage-deck-grid` here) and bubbles to `window`. Measured order on an interrupted drag: `lost@21242`, `pointerup@21243` — both only at release.
So the end signal now comes from `document` and `window`; `pointermove` stays on the node so a detached divider stops tracking. Verified: interrupted drag → flag cleared, `user-select: auto`, `cursor: auto`. Normal drag order is the reverse (`pointerup@83625`, `lost@83632`), so the final position is still persisted — re-checked with storage cleared: `session=555`, `session-analyst-chat=320`, `browser=437`.

**Namespace width keys to prevent companion ID collisions — DECLINE**
Two claims, and neither is this PR's defect.
Sharing one width across operation kinds that contribute the same companion id is the **designed behavior**, not a collision: the approved direction was per-panel-kind memory, mirroring how the right rail remembers a width per tool rather than per Operation. Making it per kind+owner would remove the property the feature was chosen for.
A plugin companion literally named `session` would indeed collide — but companion ids are already a flat namespace that this PR only consumes: `key={companion.id}` on the companion frames and the `hiddenCompanionPanelIds` / visibility-override map are both keyed by the bare id today. A duplicate id already breaks React reconciliation and panel visibility before any width is read, so renaming the width keys would not fix it, and the real fix is a companion-id uniqueness rule outside this PR's scope. Reaching it also requires authoring a plugin that picks that exact id; the two shipped companions are `session-analyst-chat` and `browser`.

Checks: `typecheck`, `test` (583 passed), `check:core-boundary`, console + desktop `build`.

sbluemin:
@codex Please review this PR.

sbluemin:
Third pass: the reported mechanism did not reproduce. `c83176aea` keeps a scoping guard anyway, for a different and disclosed reason.

**Filter resize completion to the captured pointer — DECLINE as stated, guard kept**

I drove real multi-touch through CDP (`Input.dispatchTouchEvent`) against the isolated Desktop and logged every pointer event. One finger drags the divider, a second finger lands elsewhere, then the second finger lifts:

```
pointerdown#8@863   finger 1 on the divider
gotpointercapture#8
pointerdown#9@420   finger 2 lands
pointerup#8@923     <- finger 1's own pointerup, immediately
lostpointercapture#8
```

Chromium ends the **first** touch pointer the instant a second finger touches down. No foreign pointer's release ever reached the handlers: every terminal event carried the drag's own id, and the width that got persisted came from `pointerup#8@923` — pointer 8's own last x, not the second pointer's 420. So the described failure (a second pointer's release computing and persisting a width from an unrelated `clientX`) cannot happen here, and it did not depend on the window-level listeners either: the same `pointerup#8` targets the captured element, so the pre-move code would have behaved identically.

That a second finger ends a one-finger divider drag is Chromium's own gesture arbitration, unchanged by this PR and present before it.

What I did keep is narrower than the suggestion and justified by my previous commit rather than by this scenario: moving the end listeners to `document`/`window` dropped the per-element scoping that pointer capture used to imply, so the handlers now state that invariant explicitly and ignore any pointer other than the one that started the drag. Three predicates, no observed behavior change. Re-verified after adding it: mouse drag `437/437/437` → `555/320/437` with three widths stored and the flag cleared, `→ →` on the second divider `320 → 368` / `437 → 389` with the session untouched, double-click back to `437/437/437`.

Checks: `typecheck`, `test` (583 passed), console + desktop `build`.

That is three passes. Unless something reproducible turns up, I am treating the review as complete and moving to the final audit.

PR #1170, opened 2026-09-20
feat(console): give the chat ledger a Skill glyph

sbluemin:
## Summary

`Skill` was missing from the chat ledger's tool-family table, so every call fell through to the `other` family and stood as a filled dot. Measured over 14 days of transcripts it is the 6th most-used built-in tool (915 calls) and 46% of all dots — a mark that is supposed to mean "a name we do not know" rather than "a name we never wrote down".

- Adds a `skill` letter to the glyph alphabet: an open book drawn as two strokes that each close one page, the shared edge doubling as the spine, so the silhouette stays distinct from `read`'s rectangle at 13px.
- Connects four more names whose existing family sentence is actually true: `ToolSearch` → search, `ReadMcpResourceTool` → fetch, `Artifact` → artifact (a letter already drawn but wired to nothing), `Monitor` → inspect.
- Names whose glyph would fit but whose sentence would lie stay as dots on purpose. `SendMessage` under delegate would read "delegated 1 task" when it actually spoke to an agent already running; `ScheduleWakeup` under plan would read "updated the plan". `other` prints the tool name, so "SendMessage 2회" is at least accurate.

Dot share of built-in calls drops from 4.3% to about 1.1%.

Scope note: this PR previously also carried a light-theme wave carrier change and the retirement of the ledger's spinning ring. Both were rolled back at the user's direction; only the glyph work remains.

## Test Plan

- [x] `vitest run tests/` — 556 passed, 8 skipped (127 files)
- [x] `tsc --noEmit` on `@dotobokuri/fleet-console`
- [x] `compile-changelog-fragments.mjs --check`
- [x] Glyph rendered at its real 13px size in both themes against the product's own stylesheet, confirming the open book stays legible and does not read as `read`'s document


sbluemin:
@codex Please review this PR.

sbluemin:
### Review disposition — previous pass (commit `a62cb4bf45`)

All three P2 findings are **MOOT after a scope rollback**, not declined on their merits.

| Finding | Disposition | Basis |
|---|---|---|
| Preserve the dark-theme sweep on job labels | MOOT | `.agent-chat-job-title` / `.agent-chat-ledge-name` never receive `agent-chat-live-text` on this head |
| Add a dark-theme carrier to live job-trail rows | MOOT | `agent-chat-live-row` no longer exists |
| Keep a live carrier in the compact work ledge | MOOT | the ledge is unchanged from `canary` |

The reviewer diagnosed the real tension correctly: with the orbit removed and the dark band intentionally transparent, any row whose text already carries aurora had no remaining carrier. That, together with the band spanning the full measure column instead of the streamed text, is why the wave and ring work was pulled from this PR at the author's direction.

This PR is now the glyph change only — `agent-glyphs.tsx`, `chat-events.ts`, `i18n/index.ts`, plus a changelog fragment. No CSS, no markup, no design-contract edits.

@codex Please review this PR.

PR #1175, opened 2026-09-21
fix(console): retain browser frames behind open menus

sbluemin:
## Summary
- Keep the last decoded browser frame visible while Console menus and overlays park the native view; restore the live page when they close.
- Cancel stale captures and bound capture/decode failures with safe parking fallback.
- Preserve annotation behavior and the Desktop protocol. Windows remote host-picker stacking is explicitly out of scope.

## Test Plan
- [x] Workspace typecheck and Console/Desktop builds.
- [x] Final focused tests: design contract and browser Desktop engine (99 passed); initial full Console suite (592 passed).
- [x] Headed macOS Electron: profile/viewport/local host menus, live restoration, rapid open-close-open, click-through prevention, annotation enter/exit, and capture-failure fallback.
- [x] Final client TypeScript check and `git diff --check`.
- [ ] Windows native verification (not available in this run).


sbluemin:
@codex Please review this PR.

sbluemin:

## Addressed
- FIX — discussion_r4062905320: wire response `close` (only when not writableEnded) to the screenshot service's existing AbortSignal; release the listener in finally and avoid failure responses after disconnect.
- Reproduced using the actual Fleet router and BrowserService over Node HTTP, with a stable mocked CDP transport: disconnect during the normal 50ms geometry stabilization previously still issued Page.captureScreenshot. After the fix that command is not issued; a normal request still returns 200 with screenshot data. This is a service/transport-boundary check, not a claim about measured native CDP queue latency.

## Validation
- Workspace typecheck passed.
- Console build passed.
- Design contract, browser Desktop engine, and screenshot handoff: 108 tests passed.
- Temporary real-router/service HTTP cancellation checks: normal response and early disconnect passed.

## Notes
- Cumulative review diff is limited to the screenshot route and remains within the frozen last-frame/menu request lifecycle scope. No Desktop protocol changes or Windows host-picker changes.

PR #1177, opened 2026-09-21
fix(console): hold the reading place when chat history expands

sbluemin:
## Summary

- Expanding "이전 대화 N · 펼치기" inserted every earlier turn above the current one without moving the scroll offset, so the viewport landed on the session's first question and the current turn was pushed out of view (measured: the current turn moved 123px → 1,084px in a 731px log, and reaching the collapse control again meant scrolling back 1,889px).
- The band is only reachable at the very top, and that is exactly the one offset where the browser's own scroll anchoring does not engage — so the jump happened on every expand, by structure rather than by accident.
- The toggle now remembers the distance to the log's bottom and recomputes the target offset from it. Recomputing (rather than adding the inserted height) avoids doubling the correction where the browser already anchored. A small peek lets the previous answer's tail cross above the current turn, and it drops to zero while following the bottom.

## Test Plan

- [x] `pnpm typecheck` (all workspaces)
- [x] `vitest run` — `chat-store.test.tsx`, `instrument-design-contract.test.ts` (103 passed)
- [x] Real app, isolated Console serving this build, Sonnet 5 session with 7 turns (log 731px, current turn 1,509px):
  - expand at offset 0 → eye moves **56px** (was 961px); current turn stays in view with the previous answer's tail peeking above
  - collapse → returns exactly to offset 0 / current turn at 123px
  - expand at offset 400 → eye moves **56px** (no double compensation against browser anchoring)
  - expand while following the bottom → eye moves **0px**, stays at the bottom, no peek flicker
  - new turn arrives → band re-collapses, log returns to the top, question seats at 123px (unchanged behavior)


sbluemin:
@codex Please review this PR.

